PR #8, opened 2025-01-11
change default process kill time

hjlarry:
when I install some plugin like `gemini`, `ollama`,  it always raise `init process exited due to long time no activity`.
and after I change the time, all packages installed success.

the time is always larger than 60s,  `gemini`:
3bdf5e6f4de6cc83b8cc9e5a7f15117

`siliconflow`:
ec5791d5f5b6a077bcb3b994d41da56

### DIFF
--- a/internal/core/plugin_manager/local_runtime/environment_python.go
+++ b/internal/core/plugin_manager/local_runtime/environment_python.go
@@ -161,9 +161,9 @@ func (p *LocalPluginRuntime) InitPythonEnvironment() error {
 				break
 			}
 
-			if time.Since(lastActiveAt) > 180*time.Second {
+			if time.Since(lastActiveAt) > time.Duration(p.pythonEnvInitTimeout)*time.Second {
 				cmd.Process.Kill()
-				err_msg.WriteString("init process exited due to long time no activity")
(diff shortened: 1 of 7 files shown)


internal/core/plugin_manager/local_runtime/environment_python.go
```diff
@@ -161,7 +161,7 @@ func (p *LocalPluginRuntime) InitPythonEnvironment() error {
 				break
 			}
 
-			if time.Since(lastActiveAt) > 60*time.Second {
+			if time.Since(lastActiveAt) > 180*time.Second {
```
Yeuoly: I prefer making it configurable, how do you like this?

hjlarry: Sure,  it will be more flexable !

kurokobo:
@hjlarry 
Thanks for the PR, should `internal/types/app/default.go` be updated?

hjlarry:
> @hjlarry Thanks for the PR, should `internal/types/app/default.go` be updated?

thanks for your suggestion, I think it should be modified too.

Yeuoly (approved):
LGTM

PR #12, opened 2025-01-17
fix self.session.delete alway success

hjlarry:
currently,  `self.session.delete` will always response `SUCCESS`,  even when the key is not exist.

In redis,  the `del` command will response how many keys were deleted,  it will never raise `redis.Nil`.


Yeuoly:
Hi, sorry for the late review, please resolve the CI errors, and I intent to add a `deleted_nums` as a result

hjlarry:
> Hi, sorry for the late review, please resolve the CI errors, and I intent to add a `deleted_nums` as a result

done, please review again

internal/core/plugin_daemon/backwards_invocation/task.go
```diff
@@ -531,7 +531,8 @@ func executeDifyInvocationStorageTask(
 			"data": "ok",
 		})
 	} else if request.Opt == dify_invocation.STORAGE_OPT_DEL {
-		if err := persistence.Delete(tenantId, pluginId.PluginID(), request.Key); err != nil {
+		_, err := persistence.Delete(tenantId, pluginId.PluginID(), request.Key)
```
Yeuoly: deletedNums can be added to https://github.com/langgenius/dify-plugin-daemon/pull/12/files#diff-24ca4e4a9a63a911450db3fcb9b2e366cb6eb4f8abee44af40f610974f95c4b9R541

hjlarry: In the plugin SDK, an exception will be raised if the response is not `data:ok`. Do we need to modify it in both places?
https://github.com/langgenius/dify-plugin-sdks/blob/8310e775335a655bc5ff356b23d37bcb28616d13/python/dify_plugin/invocations/storage.py#L46

Yeuoly: We can return `{"data":"ok", "deleted_nums":xx}`

Yeuoly (approved):
LGTM

PR #13, opened 2025-01-21
change hint of input uppercase letter in plugin name

hjlarry:
The hint state:
```
Bundle name must be 1-128 characters long, and can only contain letters, numbers, dashes and underscores
```

However, uppercase letters are currently not allowed.


Yeuoly:
Sorry for the confusion, but lowercase is our basic design, it's related to marketplace and Github, as we known, github username is not case-sensitive and a mixed case usage can make the namespace look messy, also, just like docker image names, it dose not allow uppercase letters.

hjlarry:
> Sorry for the confusion, but lowercase is our basic design, it's related to marketplace and Github, as we known, github username is not case-sensitive and a mixed case usage can make the namespace look messy, also, just like docker image names, it dose not allow uppercase letters.

It make sense,  so I change this PR to modify the hints

PR #14, opened 2025-01-21
feat: installation script

MatriQ:
Add install.sh that simplified installation process for the `dify-plugin-daemon` with proper distribution.


Yeuoly:
I would like to move those scripts to .scripts folder, how do you like it?

Yeuoly:
And please leave README.md empty as we will complete it at formal release of Dify v1.0

PR #17, opened 2025-01-24
fix disable endpoint not work

hjlarry:
currently, disable endpoint not work.

3e6bf202d437010a3915d3078ae9a68


Yeuoly:
Ideal, it should be set to false at internal/service/install_service/state.go#L178, I found there is a conflict between internal/service/install_service/state.go#L172 and internal/service/endpoint.go#L223, remove the internal/service/endpoint.go#L223 could make sense I guess, but it I prefer refactoring it, just using a transaction to avoid race condition by adding a mutex.

hjlarry:
I initially overlooked the `install_service.DisabledEndpoint` at first, please review it again

Yeuoly (approved):
LGTM

PR #21, opened 2025-01-29
support REDIS_USE_SSL config parameter

### LINKED ISSUE #20: support Redis TLS connection
Hi! I'm trying Dify v1 and found some issue on our infrastructure setup.
 
Currently when a Redis server only accepts TLS connection, the plugin daemon will not work because it cannot connect to redis.

We can change the code here following this doc: [Go Redis - Use TLS](https://redis.uptrace.dev/guide/go-redis.html#using-tls)

https://github.com/langgenius/dify-plugin-daemon/blob/0f19a092ef11a922916aa48df43b1b635809e406/internal/utils/cache/redis.go#L21-L25

Dify API has a config parameter called [`REDIS_USE_SSL`](https://github.com/langgenius/dify/blob/d44882c1b5aa8ec59c73d5393ef677024a5e7e1a/api/configs/middleware/cache/redis_config.py#L37-L40).

Thank you for creating Dify anyway!

tmokmss:
closes #20 

E2E-tested on my environment.

### DIFF
--- a/internal/cluster/clutser_test.go
+++ b/internal/cluster/clutser_test.go
@@ -11,7 +11,7 @@ import (
 )
 
 func createSimulationCluster(nums int) ([]*Cluster, error) {
-	err := cache.InitRedisClient("0.0.0.0:6379", "difyai123456")
+	err := cache.InitRedisClient("0.0.0.0:6379", "difyai123456", false)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/core/persistence/persistence_test.go
+++ b/internal/core/persistence/persistence_test.go
@@ -12,7 +12,7 @@ import (
 )
 
 func TestPersistenceStoreAndLoad(t *testing.T) {
-	err := cache.InitRedisClient("localhost:6379", "difyai123456")
+	err := cache.InitRedisClient("localhost:6379", "difyai123456", false)
 	if err != nil {
 		t.Fatalf("Failed to init redis client: %v", err)
 	}
@@ -72,7 +72,7 @@ func TestPersistenceStoreAndLoad(t *testing.T) {
 }
 
 func TestPersistenceSaveAndLoadWithLongKey(t *testing.T) {
-	err := cache.InitRedisClient("localhost:6379", "difyai123456")
+	err := cache.InitRedisClient("localhost:6379", "difyai123456", false)
 	if err != nil {
 		t.Fatalf("Failed to init redis client: %v", err)
 	}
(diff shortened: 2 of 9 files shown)


Yeuoly:
Great! And I wonder it could be better if there is a unittest, would you mind add some for it?

Yeuoly:
And it seems that tests are failed

tmokmss:
@Yeuoly Thanks for the review! I fixed the existing tests.

I'm afraid writing a unit test for this change can be troublesome. Do we need to spin up a Redis server with TLS setting and check if we can successfully connect to it? I think that is what go-redis already guarantees. Let me know if you have any idea 🙏

Yeuoly:
> @Yeuoly Thanks for the review! I fixed the existing tests.
> 
> I'm afraid writing a unit test for this change can be troublesome. Do we need to spin up a Redis server with TLS setting and check if we can successfully connect to it? I think that is what go-redis already guarantees. Let me know if you have any idea 🙏

Aha, it's not required, but just like some tests for configurations and environments, it could be useful to avoid issues during refactors.

Yeuoly:
> > @Yeuoly Thanks for the review! I fixed the existing tests.
> > I'm afraid writing a unit test for this change can be troublesome. Do we need to spin up a Redis server with TLS setting and check if we can successfully connect to it? I think that is what go-redis already guarantees. Let me know if you have any idea 🙏
> 
> Aha, it's not required, but just like some tests for configurations and environments, it could be useful to avoid issues during refactoring.

tmokmss:
@Yeuoly Thanks. That makes sense. I'll add one. Do you know if there are any similar existing tests for reference?

Yeuoly:
> @Yeuoly Thanks. That makes sense. I'll add one. Do you know if there are any similar existing tests for reference?

there is it, https://github.com/langgenius/dify-plugin-daemon/blob/0f19a092ef11a922916aa48df43b1b635809e406/.github/workflows/tests.yml#L20, just add a new service with SSL could make it works I guess, but a self-signed license is needed,  then add a test here https://github.com/langgenius/dify-plugin-daemon/blob/0f19a092ef11a922916aa48df43b1b635809e406/internal/utils/cache/redis_test.go#L21

tmokmss:
@Yeuoly omg sorry I misunderstood your comment and wrote another test. f38908d
If that is not required it'd be great if we could omit the integration test with real redis server, since I don't have full access to the github actions and the trial and error cycle will require more effort 😓

Yeuoly:
> @Yeuoly omg sorry I misunderstood your comment and wrote another test. [f38908d](https://github.com/langgenius/dify-plugin-daemon/commit/f38908d3fe56183639e08685051bff246f1a4c62) If that is not required it'd be great if we could omit the integration test with real redis server, since I don't have full access to the github actions and the trial and error cycle will require more effort 😓

I agree, current test you've added also make sense, let's use the simply one

Yeuoly (approved):
LGTM

PR #23, opened 2025-01-31
support AWS credentials from IAM role

tmokmss:
Currently only static IAM credentials are supported to access S3 buckets. It would be great if we could use IAM role credentials when running Dify on AWS, just [like Dify itself](https://github.com/langgenius/dify/blob/56e15d09a9e94de963dd5fc38c7455a7aaee99f3/api/extensions/storage/aws_s3_storage.py#L18-L36). This PR enables it.

I found adding unit test for this difficult but let me know if you think it is required 🙏  At least I E2E-tested on my AWS environment.

### DIFF
--- a/internal/oss/s3/s3_storage.go
+++ b/internal/oss/s3/s3_storage.go
@@ -21,20 +21,30 @@ type AWSS3Storage struct {
 }
 
 func NewAWSS3Storage(ak string, sk string, region string, bucket string) (oss.OSS, error) {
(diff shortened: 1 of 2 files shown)


Yeuoly (approved):
LGTM

Yeuoly:
Sorry for the late review, we've been in our vocation for few weeks

tmokmss:
No worries, thank you so much!

PR #26, opened 2025-02-13
fix: add missing import for llm template

euxx:
`LargeLanguageModel` is used but has not been imported.

```py
class {{ .PluginName | SnakeToCamel }}LargeLanguageModel(LargeLanguageModel):
```


Yeuoly (approved):
LGTM

PR #30, opened 2025-02-18
fix environment variables are not propagated to local runtime

### LINKED ISSUE #29: Cannot use IAM role credentials from plugin runtime
When dify-web calls API `POST console/api/workspaces/current/model-providers/langgenius/bedrock/bedrock` (by clicking Save button below), it results in an unknown error.

Image

The reason seems the python process in the plugin-daemon fails to authenticate with AWS APIs using IAM role credentials. These credentials are available via environment variables and the plugin-daemon process itself can access them (c.f. #23), but the local runtime process cannot access them because the environment variables are not propagated.

The behavior results from this code:

https://github.com/langgenius/dify-plugin-daemon/blob/0636968f84e86191af40e1e0ad331abed23f23e7/internal/core/plugin_manager/local_runtime/run.go#L77-L84

`e.Env = append(e.Env,` effectively erases all the environment variables because e.Env is nil by default. I'm not sure if it is expected or not, but if we want to propagate environment variables, we can write it like the below [c.f. in this document](https://pkg.go.dev/os/exec#Command).

```go
	// Set Dir before calling cmd.Environ so that it will include an
	// updated PWD variable (on platforms where that is used).
	cmd.Dir = ".."
	cmd.Env = append(cmd.Environ(), "POSIXLY_CORRECT=1")
```

tmokmss:
closes #29 

If it is the expected behavior to pass environment variables to local runtime process (not sure about this, but at least it is desired for self-host environment), this should be the fix.

### DIFF
--- a/internal/core/plugin_manager/local_runtime/run.go
+++ b/internal/core/plugin_manager/local_runtime/run.go
@@ -36,6 +36,7 @@ func (r *LocalPluginRuntime) getCmd() (*exec.Cmd, error) {
 	if r.Config.Meta.Runner.Language == constants.Python {
 		cmd := exec.Command(r.pythonInterpreterPath, "-m", r.Config.Meta.Runner.Entrypoint)
 		cmd.Dir = r.State.WorkingPath
+		cmd.Env = cmd.Environ()
 		if r.HttpsProxy != "" {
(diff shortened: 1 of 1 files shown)


Yeuoly (approved):
LGTM

PR #49, opened 2025-03-04
Add missing `url` field of PromptMessageContent

chekun:
MultiModal content supports both base64 and url format, but lost in dify-plugin-daemon server


Yeuoly (approved):

Yeuoly:
Thanks so much for the contribution! and so sorry for the bugs, we are focusing on fix them

PR #59, opened 2025-03-05
fix: optimize plugin last active time tracking in stdio handler

Yeuoly:
Move last active time update to occur on each data receipt to ensure more accurate tracking of plugin activity


joebnb:
is there need ship a version for user update

PR #61, opened 2025-03-05
fix: Fix CROSSSLOT Error in Redis Commands

Jam804:
This pull request addresses the issue of the "CROSSSLOT Keys in request don't hash to the same slot" error, which occurs when Redis commands involve keys that are not hashed to the same slot in a clustered environment.

Changes Made:

1. Updated Redis command logic to ensure that all keys involved in transactions or multi-key operations are hashed to the same slot.
2. Modified the data distribution strategy to align with Redis cluster requirements, ensuring key patterns are consistent with slot allocation.

### DIFF
--- a/internal/core/plugin_manager/debugging_runtime/connection_key.go
+++ b/internal/core/plugin_manager/debugging_runtime/connection_key.go
@@ -29,8 +29,8 @@ type Key struct {
 }
 
 const (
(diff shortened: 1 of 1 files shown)


Yeuoly (approved):
LGTM, thanks for the contribution!

PR #69, opened 2025-03-06
fix: Fix issues in the example configuration file.

Jam804:
This configuration item has extra quotes added.


perzeuss:
Why do you think this is wrong? These are double quotes and not extra single quotes.
It's okay to double quote values.

PR #52, opened 2025-03-04
fix fatal error: concurrent map iteration and map write

qkzsky:
fix
fatal error: concurrent map iteration and map write

PR #72, opened 2025-03-07
fix: prevent potential deadlock in stdio handler listener invocation

Yeuoly:
Modify stdout listener handling to avoid potential deadlocks by:
- Copying listeners before invocation
- Executing listener callbacks outside of the lock
- Separating listener retrieval and execution steps


Yeuoly:
FIXES: https://github.com/langgenius/dify-plugin-daemon/pull/52

PR #74, opened 2025-03-07
fix tts not work well

hjlarry:
fix https://github.com/langgenius/dify/issues/15025

I add this code to plugin sdk [here](https://github.com/langgenius/dify-plugin-sdks/blob/b936b1be9f91d1ae6725604ad8938b80cf42c635/python/dify_plugin/core/plugin_executor.py#L217) for debug:
```
            for i, chunk in enumerate(b):
                with open(f'tts_debug_chunk_{i}.bin', 'wb') as f:
                    f.write(chunk)
                yield {"result": binascii.hexlify(chunk).decode()}
```

also add this code to api server [here](https://github.com/langgenius/dify/blob/4aaf07d62af180439a4871a2191e45b610929881/api/core/plugin/manager/model.py#L403) for debug:
```
            for i, result in enumerate(response):
                hex_str = result.result
                decoded = binascii.unhexlify(hex_str)
                with open(f'received_chunk_{i}.bin', 'wb') as f:
                    f.write(decoded)
                yield decoded
```

always get  `tts_debug_chunk_70.bin` files  and only `received_chunk_50.bin` files
after change the `response_buffer_size` it works

### DIFF
--- a/internal/core/plugin_daemon/model_service.go
+++ b/internal/core/plugin_daemon/model_service.go
@@ -55,7 +55,7 @@ func InvokeTTS(
 	return GenericInvokePlugin[requests.RequestInvokeTTS, model_entities.TTSResult](
 		session,
 		request,
-		1,
+		512,
 	)
 }
 


Yeuoly (approved):
LGTM

Issue #102, opened 2025-03-12
Application fails to start when HTTP_PROXY is set - “unexpected argument ‘–proxy’ found”


varluffy:
Steps to Reproduce:
1. Set the HTTP_PROXY environment variable
2. Start the application.
3. The application attempts to install dependencies and fails with the above error.

```shell
runtime_lifetime.go:76: [ERROR]init environment failed: failed to install dependencies: exit status 2, output: error: unexpected argument '--proxy' found
```

kurokobo:
Fixed in #99

kurokobo:
@Yeuoly 
I've just sent follow-up PR for your #99: https://github.com/langgenius/dify-plugin-daemon/pull/119

PR #119, opened 2025-03-18
fix: pass HTTP_PROXY and HTTPS_PROXY to uv process

### LINKED ISSUE #118: HTTP_PROXY and HTTPS_PROXY are not used during "uv pip install"
In the environment under corporate proxy, the plugin daemon should have `HTTP_PROXY` and `HTTPS_PROXY` environment variables.

The invalid argument `–proxy` (issue https://github.com/langgenius/dify-plugin-daemon/issues/102) has been removed by #99; however, neither `HTTP_PROXY` nor `HTTPS_PROXY` has been passed to `cmd`.

https://github.com/langgenius/dify-plugin-daemon/blob/84edb2682d569dca2011564d4d1c13f0a42fdc9f/internal/core/plugin_manager/local_runtime/environment_python.go#L124-L125

This is causing the failure of the dependency installation due to inaccessibility to PyPI. The changes by #99 were correct to suppress argument errors, but they were insufficient to ensure proper behavior in a proxy environment.

kurokobo:
This PR changes `uv` command to use proxy if it is specified by `HTTP_PROXY` and/or `HTTPS_PROXY`.

This is follow-up for #99 for #102, also closes #118

### DIFF
--- a/internal/core/plugin_manager/local_runtime/environment_python.go
+++ b/internal/core/plugin_manager/local_runtime/environment_python.go
@@ -123,6 +123,12 @@ func (p *LocalPluginRuntime) InitPythonEnvironment() error {
 	virtualEnvPath := path.Join(p.State.WorkingPath, ".venv")
 	cmd = exec.CommandContext(ctx, uvPath, args...)
 	cmd.Env = append(cmd.Env, "VIRTUAL_ENV="+virtualEnvPath, "PATH="+os.Getenv("PATH"))
+	if p.HttpProxy != "" {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("HTTP_PROXY=%s", p.HttpProxy))
+	}
+	if p.HttpsProxy != "" {
+		cmd.Env = append(cmd.Env, fmt.Sprintf("HTTPS_PROXY=%s", p.HttpsProxy))
+	}
 	cmd.Dir = p.State.WorkingPath
 
 	// get stdout and stderr


Yeuoly (commented):
LGTM

Yeuoly:
It's weird why HTTP_PROXY and HTTPS_PROXY are not passed to env by default,

kurokobo:
@Yeuoly 
This is expected behavior of `os/exec`.

Quick example: https://go.dev/play/p/OxlcX3f__hZ

- **cmd2** is the same way as it is for `uv pip install`
- **cmd3** is the same way as it is for `python -m main`

The key points are:

- Whether `cmd.Env` is explicitly configured.
- Whether `cmd.Environ()` is used or not to configure `cmd.Env`.

This PR didn't make any modifications, but in fact, considering this, the code here is a bit confusing:

https://github.com/langgenius/dify-plugin-daemon/blob/a0e595c3b28af79aa5cd52ddca2a2b7b6e886ab1/internal/core/plugin_manager/local_runtime/run.go#L39-L45

Since the `cmd.Env` is initialized using `cmd.Environ()`, there is no need to explicitly set `HTTP_PROXY` and `HTTPS_PROXY`.

On the contrary, because it is initialized with `cmd.Environ()`, plugins operating in `local` mode can access all environment variables held by the daemon, including sensitive information like database credentials. I think this is what was intended with #30.

Yeuoly:
> @Yeuoly This is expected behavior of `os/exec`.
> 
> Quick example: https://go.dev/play/p/OxlcX3f__hZ
> 
> * **cmd2** is the same way as it is for `uv pip install`
> * **cmd3** is the same way as it is for `python -m main`
> 
> The key points are:
> 
> * Whether `cmd.Env` is explicitly configured.
> * Whether `cmd.Environ()` is used or not to configure `cmd.Env`.
> 
> This PR didn't make any modifications, but in fact, considering this, the code here is a bit confusing:
> 
> https://github.com/langgenius/dify-plugin-daemon/blob/a0e595c3b28af79aa5cd52ddca2a2b7b6e886ab1/internal/core/plugin_manager/local_runtime/run.go#L39-L45
> 
> Since the `cmd.Env` is initialized using `cmd.Environ()`, there is no need to explicitly set `HTTP_PROXY` and `HTTPS_PROXY`.
> 
> On the contrary, because it is initialized with `cmd.Environ()`, plugins operating in `local` mode can access all environment variables held by the daemon, including sensitive information like database credentials. I think this is what was intended with #30.

I have tried it before, without this, subprocesses are leak of necessary environments, for now, the most easiest way to make it safety was `signature`, but, yeah, I agree it's in danger

PR #122, opened 2025-03-19
Specifies the SSL mode for the database connection

ke-feng:
Description
-----------------
Specifies the SSL mode for the database connection explicitly in the .env. 
Default value is disable.


Yeuoly (approved):
LGTM

PR #125, opened 2025-03-20
add gorm easer cache policy

### LINKED ISSUE #124: call endpoint concurrency result in 500
plugin daemon log:

```
plugin_daemon-1  | 2025/03/20 04:28:54 /app/internal/db/pgsql.go:259 failed to connect to `host=db user=postgres database=dify_plugin`: server error (FATAL: sorry, too many clients already (SQLSTATE 53300))
```

zxfishhack:
gorm easer cache policy will merge same query and execute once as same time, this will solve 'too many clients already' when concurrency access plugin endpoint.

Closed #124

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,9 +13,10 @@ require (
 	github.com/charmbracelet/bubbletea v1.1.0
 	github.com/fxamacker/cbor/v2 v2.7.0
 	github.com/getsentry/sentry-go v0.30.0
-	github.com/gin-contrib/gzip v1.0.1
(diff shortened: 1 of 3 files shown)


Yeuoly:
Hi, would you mind add some tests for this plugin? and how do it keep the consistency, for example, in a transaction, will queries be considered as `the same time`? I'm not used to it before

zxfishhack:
That's a problem, in a transaction, query will hit cache, and the queries be sonsidered same as other queries not in transaction.

zxfishhack:
I'll use custom cache solve it.

PR #128, opened 2025-03-20
add cache for endpoint by hook_id and pluginInstallation by plugin_id&tenant_id

### LINKED ISSUE #124: call endpoint concurrency result in 500

zxfishhack:
Add cache for endpoint and pluginInstallation to reduce database query in App.EndpointHandler.

Closed #124 

**Note**: The modifications of this PR are relatively scattered, and the invalidation of cache is scattered in the disabling and deletion of Endpoint, uninstallation of plugins, etc. If there is a plan to maintain the cache in a unified manner, I will wait for it.


Yeuoly:
how do you like `AutoGetWithGetter` in `internal.utils.cache.redis_auto_type.go` ?

zxfishhack:
It looks like an alternative implementation, but it uses a custom cache key.
If the project uses the method here in the future, I will switch to this implementation.

BTW: I think that when writing cache fails, the error should not be returned. The data obtained from the database should be returned to the upper layer, and a log of cache write exception should be generated.

Yeuoly:
> It looks like an alternative implementation, but it uses a custom cache key. If the project uses the method here in the future, I will switch to this implementation.
> 
> BTW: I think that when writing cache fails, the error should not be returned. The data obtained from the database should be returned to the upper layer, and a log of cache write exception should be generated.
> 
> In addition, is there a unified place to manage the generation method of cache key?

Hmmm, it's my miss, I forgot to delete the method you are using, and for custom key, AutoGetWithGetter uses reflect to generate prefix key, for `Endpoint`, the prefix will be `langgenius/dify-plugin-daemon/internal/models/endpoint.Endpoint

zxfishhack:
I've tried using `AutoGetWithGetter`, but the cache is not hit using this method.

In function `AutoGetWithGetter`:
```go
			if err := Store(key, value, time.Minute*30, context...); err != nil {
```
`Store` will call `serialKey` again, The key used for storage is inconsistent with the key used for retrieval.

Should I correct this in this PR or a new one?

zxfishhack:
I have open a new PR about the cache issue, I think add an internal function named `store` is better, The `AutoType` series of cache implementations will manage the construction of keys themselves. If `Get[T]` is called when obtaining, it must also be modified when deleting.

#140

Yeuoly:
> I have open a new PR about the cache issue, I think add an internal function named `store` is better, The `AutoType` series of cache implementations will manage the construction of keys themselves. If `Get[T]` is called when obtaining, it must also be modified when deleting.
> 
> #140

Agree, and `get`/`delete` implementations are also needed, just for internal uses without serialKey

zxfishhack:
The implementation has been adjusted to use cache.AutoGetWithGetter instead.

internal/service/install_service/state.go
```diff
@@ -134,6 +137,15 @@ func GetEndpoint(
...
 // uninstalls a plugin from db
 func UninstallEndpoint(endpoint *models.Endpoint) error {
+	cacheKey := gostrings.Join(
+		[]string{
+			"hook_id",
+			endpoint.HookID,
+		},
+		":",
+	)
+	log.Print(cacheKey)
```
Yeuoly: this line should be removed

Yeuoly (changes requested):

Yeuoly:
Would you mind add some tests for this? as we are building a robust testing pipeline

internal/service/install_service/state.go
```diff
@@ -202,6 +214,18 @@ func DisabledEndpoint(endpoint_id string, tenant_id string) error {
...
 
+		endpointCacheKey := gostrings.Join(
+			[]string{
+				"hook_id",
+				endpoint.HookID,
+			},
+			":",
+		)
+
+		log.Print(endpointCacheKey)
```
Yeuoly: And this line

zxfishhack:
> Would you mind add some tests for this? as we are building a robust testing pipeline

Is there a simple way to create plugin daemon environment in test?

Yeuoly:
> > Would you mind add some tests for this? as we are building a robust testing pipeline
> 
> Is there a simple way to create plugin daemon environment in test?

I guess what you mention is a integration test, unfortunately only unit tests are supported, but a redis instance already exists in CI pipeline, checkout https://github.com/langgenius/dify-plugin-daemon/blob/main/internal/utils/cache/redis_auto_type_test.go#L13

zxfishhack:
A unittest for `AutoGetWithGetter` had already in PR #140 by modify `TestAutoTypeWithGetter`, it will try to call `AutoGetWithGetter` twice, first with a normal getter, second with a getter alway return an error, which confirm `AutoGetWithGetter` will hit cache.

IMO, to test this PR, the following points need to be implemented:
1. Whether the cache will perform corresponding operations with the creation, disabling, and deletion of the endpoint
2. Whether the cache will perform corresponding operations with the deletion of the plugin
3. Whether the cache can be hit when called multiple times

Yeuoly:
Hi, pls resolve the CI errors

zxfishhack:
> Hi, pls resolve the CI errors

Done by merged main, the cache test error has been solved by another PR.

Yeuoly (approved):
LGTM

Issue #134, opened 2025-03-22
s3 compatible storage config should support configuring pathStyle


AkisAya:
![Image](https://github.com/user-attachments/assets/1ec7f63f-47c0-435f-8ee2-ce320f6f5b15)

now pathStyle is hard coded to true. however some s3 compatible storage service need this to be false(which means using a virtual host)

Yeuoly:
Thanks!

PR #135, opened 2025-03-22
make pathStyle param configurable when using s3 compatible storage

AkisAya:
#134 
make pathSytle param of s3 compatible storage services configurable


AkisAya:
S3_BUCKET_NAME is introduced but never used, actually PluginStorageOSSBucket is used to configure s3 bucket.
this pr also removes S3_BUCKET_NAME in env.example.
and some chore updates about logs

Yeuoly (approved):
LGTM

PR #136, opened 2025-03-23
support to use mysql as plugin db

whhe:
In this PR, I added a new module `internal/db/mysql`, which implements the `initDifyPluginDB` of MySQL. Models definitions that are incompatible with MySQL (such as uuid types and text fields) are rewritten through custom dialector and migrator. These changes will not affect the original postgresql process. PTAL @Yeuoly


Yeuoly:
Hi, do you have any idea on long term maintenance of mysql implementation? It's hard to manage two dialectors at the same time

whhe:
> Hi, do you have any idea on long term maintenance of mysql implementation? It's hard to manage two dialectors at the same time

Hi @Yeuoly, thanks for your reply. 

At present, the database-related processes are already generic enough, and theoretically there won't be many new incompatibilities. I think we can use mysql as an optional solution in short term, and mark it as production-ready after a long term verification.

Our team has just created a MySQL-compatible Dify fork and started using it internally, and we will follow up with the official releases in the future. I think we can act as a user and tester for this feature, and fix related bugs in time. What do you think?

internal/types/models/base.go
```diff
@@ -2,10 +2,20 @@ package models
...
 
 type Model struct {
 	ID        string    `gorm:"column:id;primaryKey;type:uuid;default:uuid_generate_v4()" json:"id"`
 	CreatedAt time.Time `json:"created_at"`
 	UpdatedAt time.Time `json:"updated_at"`
 }
+
+func (m *Model) BeforeCreate(tx *gorm.DB) (err error) {
+	if tx.Dialector.Name() == "mysql" && m.ID == "" {
+		m.ID = uuid.New().String()
```
Yeuoly, line 18: For example, the ID generator depends on dialector, not it's tag, I'm not sure will there a change to use asc integer id generator, and `default:uuid_generate_v4()` is indeed a postgres function, I'm think about how to uniform them

whhe: Like removing the default value in the gorm tag of the base model and always generating uuid via the golang package here? I chose not to change the model definition as much as possible, but in theory this implementation should not cause any problems.

Yeuoly: I agree for now it works well, but it's just a little confusing, mysql/pgsql were handed in different way, cloud you pls add some tests for it on CI pipeline? just to keep the correctness and avoid breaking changes

whhe: I added mysql config to `TestLaunchAndClosePluginServer`, it should cover the initialization of table models.

Yeuoly (approved):
LGTM

Yeuoly:
Thanks for your contributions!

PR #140, opened 2025-03-24
BUGFIX: fix AutoGetWithGetter will never hit cache issue.

zxfishhack:
add a new function named `store` for internal use, without calling serialKey.

modify `TestAutoTypeWithGetter` to call twice, check if cache hit.

### DIFF
--- a/internal/utils/cache/redis_auto_type_test.go
+++ b/internal/utils/cache/redis_auto_type_test.go
@@ -35,7 +35,7 @@ func TestAutoType(t *testing.T) {
 }
 
 func TestAutoTypeWithGetter(t *testing.T) {
-	if err := InitRedisClient("192.168.125.241:26379", "difyai123456", false); err != nil {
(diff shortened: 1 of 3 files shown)


internal/utils/cache/redis_auto_type_test.go
```diff
@@ -32,7 +35,7 @@ func TestAutoType(t *testing.T) {
 }
 
 func TestAutoTypeWithGetter(t *testing.T) {
-	if err := InitRedisClient("127.0.0.1:6379", "difyai123456", false); err != nil {
+	if err := InitRedisClient("192.168.125.241:26379", "difyai123456", false); err != nil {
```
Yeuoly: pls change to 127.0.0.1 instead

zxfishhack: oops, it's my mistake.

Yeuoly (approved):
LGTM

PR #155, opened 2025-03-27
Add get/delete implement, for internal use.

zxfishhack:
Add get/delete implement, without `serialKey`, for internal use, as metioned in #128


Yeuoly (approved):
Thanks~ LGTM

PR #160, opened 2025-03-28
fix: thread safety issue and len inaccuracy in Map

lengyhua:
**Thread-Safety Issue:**

> Even if atomic operations (e.g., atomic.AddInt32) and mutexes (e.g., sync.RWMutex) are used, modifying multiple variables (e.g., store and len) in a single method creates a critical section. For example:
> In Store, the sequence Load → AddInt32 → Store is not atomic as a whole, leading to potential race conditions.
> If two goroutines simultaneously call Store for the same key, both might see loaded = false and increment len, resulting in an incorrect count.
> 

**len Inaccuracy:**

> The len field is intended to track the number of unique keys in the map.
> In the original Store method, len was incremented unconditionally, even when overwriting an existing key. This caused len to overcount entries.

```
// Goroutine 1
m.Store("key", 1)  // loaded = false → len++ → store.Store

// Goroutine 2
m.Store("key", 2)  // loaded = false → len++ → store.Store

// Result:
// len = 2 (incorrect), but the map contains only 1 entry ("key").
```

### DIFF
--- a/internal/utils/mapping/sync.go
+++ b/internal/utils/mapping/sync.go
@@ -8,9 +8,12 @@ import (
 type Map[K comparable, V any] struct {
 	len   int32
 	store sync.Map
+	mu    sync.RWMutex
 }
 
 func (m *Map[K, V]) Load(key K) (value V, ok bool) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
 	v, ok := m.store.Load(key)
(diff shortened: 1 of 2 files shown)


Yeuoly:
Great!

Yeuoly (approved):
LGTM

PR #162, opened 2025-03-30
fix: some time uv venv wrong python version

zalcit:
The plugin is developed for Python 3.12. However, when the system Python version differs (e.g., 3.13), the plugin may fail to run correctly.

To ensure compatibility, we'll configure uv to default to Python 3.12, matching the manifest specification.


Yeuoly (approved):
LGTM

PR #170, opened 2025-04-02
Increase NODE_DISCONNECTED_TIMEOUT

liuzhenghua:
The current value of NODE_DISCONNECTED_TIMEOUT is too small. When deleting and creating a new pod, tasks like tickerLockMaster and tickerUpdateNodeStatus may block the tickerUpdateNodeStatus task, causing the health node to be marked as not available, which results in all plugin states being removed.

efc5b9e351d22be82ed84328a19e76c

### DIFF
--- a/internal/cluster/vote.go
+++ b/internal/cluster/vote.go
@@ -51,7 +51,7 @@ func (c *Cluster) voteAddresses() error {
 				}
 			}
 
-			ipsVoting[addr.fullAddress()] = c.voteAddress(addr) == nil
+			ipsVoting[addr.fullAddress()] = time.Since(time.Unix(nodeStatus.LastPingAt, 0)) < c.nodeDisconnectedTimeout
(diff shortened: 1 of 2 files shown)


internal/cluster/vote.go
```diff
@@ -51,7 +51,7 @@ func (c *Cluster) voteAddresses() error {
 				}
 			}
 
-			ipsVoting[addr.fullAddress()] = c.voteAddress(addr) == nil
+			ipsVoting[addr.fullAddress()] = time.Since(time.Unix(nodeStatus.LastPingAt, 0)) < c.nodeDisconnectedTimeout
```
fatelei, line 54: here，address can be myself，so check `time.Since(time.Unix(nodeStatus.LastPingAt, 0)) < c.nodeDisconnectedTimeout` will always be true

PR #172, opened 2025-04-03
feat(storage): add azure blob storage support

te-chan:
Hi, I’ve added Azure Blob Storage support.
I tested this storage type locally with Dify version 1.1.3.

There were no issues with adding, removing, or listing plugins.
Also, everything worked correctly even after restarting or deleting the container and volume.

I can update dify-docs if needed.


.env.example
```diff
@@ -23,6 +23,10 @@ TENCENT_COS_SECRET_KEY=
 TENCENT_COS_SECRET_ID=
 TENCENT_COS_REGION=
 
+# azure blob storage credentials
+AZURE_BLOB_STORAGE_CONTAINER_NAME=
+AzureBlobStorageConnectionString=
```
Yeuoly: Please use the same style naming convention

Yeuoly (changes requested):

.env.example
te-chan: I've updated it to use same variable naming style

te-chan:
Thank you for reviewing. I fixed variable name.

Yeuoly (approved):
LGTM

PR #182, opened 2025-04-08
fix UV_CACHE_DIR not worker

qkzsky:
fix UV_CACHE_DIR not worker


qkzsky:
In the usage environment of K8S, sometimes the hard disk space is controlled. At this time, it is necessary to customize the cache storage address of UV to the mounted directory. Currently, the configuration of environment variables is not effective. When running UV, the information of environment variables is not obtained. It is expected that when UV is executed, the environment variables can be directly used, such as UV CACHE DIR, etc

PR #187, opened 2025-04-09
add uv path config

hjlarry:
For the local development environment, `uv` is typically installed using `pipx`, so it cannot be located by `find_uv_bin`.


Yeuoly:
Can we give it a default value in docker deployment? maybe a little performance optimization.

hjlarry:
> Can we give it a default value in docker deployment? maybe a little performance optimization.

done

Yeuoly (approved):

PR #195, opened 2025-04-11
add `exist` for self.session.storage

hjlarry:
Currently,  get or delete a non-existent key will raises an exception, although this is a common operation. 

On the other hand, we can use the storage as a pipeline, frequently retrieving data from it. This results in numerous error logs on the daemon. Using 'exists' can significantly reduce these error logs.


Yeuoly (approved):
LGTM

PR #210, opened 2025-04-16
fix(plugin_manager): skip hidden files in InstalledBucket List method

Yeuoly:


Yeuoly:
cc @bowenliang123

bowenliang123:
LGTM. Very helpful in most case for macOS users. Thanks for the rapid response.

Issue #133, opened 2025-03-22
PluginDaemonBadRequestError when downloading/uploading local plugins on Windows 11(Solved!)


mymian:
Environment:
   Operating System: Windows 11
   Go Version: 1.24.1
   dify-plugin-daemon: latest
   dify1.1.1
Steps to Reproduce:
   Create a .env file in the cmd/serve directory.
   Run the command go run main.go in the terminal.
   Successfully start the Dify project locally.
   Attempt to download or upload a local plugin.

![Image](https://github.com/user-attachments/assets/d4cab432-238d-4259-b09d-1939c7b7f97a)

Expected Result:
   The plugin should be downloaded/uploaded successfully without any errors.
Actual Result:
   The following error occurs:
   PluginDaemonBadRequestError: file not found: icon.svg failed to remap tool icon failed to remap assets failed to save package

![Image](https://github.com/user-attachments/assets/141743ac-d57a-47e2-9cf2-dee2e0a32358)

![Image](https://github.com/user-attachments/assets/bcd8ef75-1f38-4eb0-8336-a8eb3a3086b2)

![Image](https://github.com/user-attachments/assets/826d41f2-7d4f-4ec3-b512-5d9bec3007c6)

![Image](https://github.com/user-attachments/assets/2f9929a4-0a7b-436d-ba4c-b5c0e300270b)

Additional Information:
   The plugin being downloaded/uploaded does contain the icon.svg file, as verified multiple times.
    I'm wondering if there is a way to resolve this issue by modifying the .env file.

![Image](https://github.com/user-attachments/assets/18f54666-0ef0-45d0-bb80-a422278af0f7)

![Image](https://github.com/user-attachments/assets/d27f3352-7b85-4871-9935-63db07d26215)

Request:
Could you please provide guidance on how to fix this error? Is there a specific configuration in the .env file that might be causing this issue?

mymian:
直接就写中文了。
在win11本地启动运行这个组件的时候踩了不少坑，最主要的问题就是双斜杠\\ 识别不了的bug，在代码中找到对应的路径将 \\ 改成 / 即可。
比如zip.go 中解压：原代码为 filename = filepath.Join(workingPath, filename) ，报错copy plugin to working directory error: open .git\COMMIT_EDITMSG: file does not exist open .git\COMMIT_EDITMSG: file does not exist extract plugin to working directory error
[...]

PR #223, opened 2025-04-17
feat: add windows compatibility (#133)

listeng:
add windows compatibility (#133)

1. Use the filepath.ToSlash() function to ensure slash normalization.
2. Temporarily replace (:) with ($) symbols in Win.
3. Force utf8 in Win to avoid garbage string.
4. Compatible paths for python executable programs in Win.

### DIFF
--- a/internal/core/plugin_manager/local_runtime/environment_python.go
+++ b/internal/core/plugin_manager/local_runtime/environment_python.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 	"sync"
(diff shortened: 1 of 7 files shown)


Yeuoly:
I intent to move all the compatible code to a generic function like `ToOsCompatPluginPath` etc, it returns `plugin_unique_identifier` depends on the specific OS, so that we can also add tests for it to ensure the robust.

As for `.venv/Script/python.exe`, we can do the same optimization, the path should be returned by another function, instead of handing it inside `StartPlugin`

internal/core/plugin_manager/local_runtime/environment_python.go
```diff
@@ -32,7 +33,11 @@ func (p *LocalPluginRuntime) InitPythonEnvironment() error {
 			os.RemoveAll(path.Join(p.State.WorkingPath, ".venv"))
 		} else {
 			// setup python interpreter path
-			pythonPath, err := filepath.Abs(path.Join(p.State.WorkingPath, ".venv/bin/python"))
+			pythonExec := ".venv/bin/python"
```
Yeuoly, line 36: this can be simplify to `p.getVenvPythonExecPath()`

internal/core/plugin_manager/media_transport/installed_bucket.go
```diff
@@ -22,28 +23,44 @@ func (b *InstalledBucket) Save(
 	plugin_unique_identifier plugin_entities.PluginUniqueIdentifier,
 	file []byte,
 ) error {
-	return b.oss.Save(filepath.Join(b.installedPath, plugin_unique_identifier.String()), file)
+	compat_plugin_unique_identifier := plugin_unique_identifier.String()
```
Yeuoly, line 26: the same code snippet repeats many times, I want it to be an util function

Yeuoly (changes requested):

PR #263, opened 2025-05-07
support concurrently handle local plugins

bowenliang123:


internal/core/plugin_manager/watcher.go
```diff
@@ -66,28 +67,43 @@ func (p *PluginManager) startRemoteWatcher(config *app.Config) {
...
+		waitGroup  sync.WaitGroup
+		semaphores = make(chan struct{}, config.PluginLocalLaunchingConcurrent)
+	)
+
 	for _, plugin := range plugins {
-		_, launchedChan, errChan, err := p.launchLocal(plugin)
-		if err != nil {
-			log.Error("launch local plugin failed: %s", err.Error())
-		}
+		waitGroup.Add(1)
```
Yeuoly: I intent to move this line outside this loop, use `waitGroup.Add(len(plugins))

Yeuoly (changes requested):

internal/core/plugin_manager/watcher.go
bowenliang123: SGTM. Resolved.

bowenliang123:
> Actually, concurrently launching was supported already, `PluginManager.maxLaunchingLock ` is what you want, it already limited concurrency of how many plugins could launch at the same time inside `launchLocal`, just feel free to launch them at the same time, and wait for all of them to finished.

Curious to hear that concurrency has been supported in launching local plugins, while in practise we don't see evidence in concurrently installed and launching local plugins.

And BTW, `config.PluginLocalLaunchingConcurrent` config is existed but never used across the project without this PR.

internal/core/plugin_manager/watcher.go
```diff
@@ -66,28 +67,43 @@ func (p *PluginManager) startRemoteWatcher(config *app.Config) {
...
+func (p *PluginManager) handleNewLocalPlugins(config *app.Config) {
 	plugins, err := p.installedBucket.List()
 	if err != nil {
 		log.Error("list installed plugins failed: %s", err.Error())
 		return
 	}
 
+	var (
+		waitGroup  sync.WaitGroup
+		semaphores = make(chan struct{}, config.PluginLocalLaunchingConcurrent)
```
Yeuoly, line 79: It's no need to use semaphores as pluginManager will automatically control its limits

Yeuoly:
Refactored: #499

PR #267, opened 2025-05-08
feat: allow skipping Python plugin precompilation

bowenliang123:
- add `SkipPrecompilation` config to allow skip Python plugin precompilation, which is handy to speed-up local plugin launching to skip overhead time cost in pre-compiling all the Python scripts and its dependency in each plugin


Yeuoly:
Refactored: #499

Issue #269, opened 2025-05-08
[feature]redis cluster&sentinel


chenzikun:
dify主服务已经支持了redis集群模式，如果这里不支持的话，其实也没有办法把redis集群模式部署起来

[相关讨论](https://github.com/langgenius/dify-plugin-daemon/issues/147)

zhuzihao-hz:
Hi👋, may I ask in which version this feature is planned to be supported?

PR #270, opened 2025-05-08
[feat] support cluster and sentinel deployment for redis

### LINKED ISSUE #147: REDIS SENTINELS
Does the current version support redis sentinel cluster, and is there any plan to support it in the future?

chenzikun:
- cluster已经过测试，sentinel暂未测试

[feature-redis cluster & sentinel](https://github.com/langgenius/dify-plugin-daemon/issues/269)

fix #147

### DIFF
--- a/pkg/utils/cache/redis.go
+++ b/pkg/utils/cache/redis.go
@@ -532,12 +532,16 @@ func Transaction(fn func(redis.Pipeliner) error, keys ...string) error {
 		return ErrDBNotInit
 	}
 
-	// 如果没有提供 keys，则返回错误
+	// Fix: If no keys provided, use plain Pipeline instead of Watch transaction
 	if len(keys) == 0 {
-		return errors.New("redis: Watch requires at least one key")
+		_, err := client.TxPipelined(ctx, fn)
+		if err == redis.Nil {
+			return nil
+		}
+		return err
 	}
 
-	// 将 keys 进行序列化处理
+	// Serialize watch keys
 	watchKeys := make([]string, len(keys))
 	for i, key := range keys {
 		watchKeys[i] = serialKey(key)


Yeuoly (changes requested):
Please use English comments instead

erigo:
see: redis.UniversalClient & https://github.com/langgenius/dify-plugin-daemon/pull/276

chenzikun:
@Yeuoly [done]  convert chinese comment to english, pls check!

Yeuoly:
I've reviewed #276, as for me, and I prefer to use redis.UniversalClient as the client so that this is no need to maintain a interface like what you do, @chenzikun @erigo , how do you like?

erigo:
I suggest adding support for Redis cluster based on #276.

AkisAya:
i think sentinel mode and cluster mode of reids subpub will not work as expected. when there is a failure of a redis instance, pubsub connection of failed redis will not automaticly change to another redis instance.  
it's better to split the cache ability to different interfaces .   https://github.com/langgenius/dify-plugin-daemon/issues

erigo:
Thanks for pointing this out — I agree with your observation that Redis Sentinel and Cluster modes do not automatically recover Pub/Sub connections after failover, which makes Redis Pub/Sub unsuitable for high-availability scenarios.

However, I would respectfully disagree with the idea of “splitting the cache ability into different interfaces.”
In practice, it's often better to unify Redis access through a single interface using redis.UniversalClient, which allows seamless switching between standalone, Sentinel, and Cluster modes. This approach simplifies the codebase, avoids redundant abstractions, and still gives us flexibility for future changes.

As for Pub/Sub, I agree that Redis is not the right tool for reliable messaging. For systems that depend on high-availability message delivery, dedicated message queues like Kafka or NATS are far more suitable.

So in short: use redis.UniversalClient to unify caching access, but avoid Redis Pub/Sub for HA-critical workflows.

AkisAya:
> use redis.UniversalClient to unify caching access
this will make it hard to extend different caching storages. e.g. our redis cluster implemention is not opensource RedisCluster or Sentinel

> but avoid Redis Pub/Sub for HA-critical workflows
agree. and it's better to split the current responsibilities of the cache into a cache interface and pubsub

Yeuoly:
> > use redis.UniversalClient to unify caching access
> 
> this will make it hard to extend different caching storages. e.g. our redis cluster implemention is not opensource RedisCluster or Sentinel
> 
> > but avoid Redis Pub/Sub for HA-critical workflows
> 
> agree. and it's better to split the current responsibilities of the cache into a cache interface and pubsub

I agree splitting them into different interfaces would make cache more extendable, but just at this point, Redis is the production standard for almost all the software, adding support for other cache is not a priority I guess

chenzikun:
@Yeuoly ok.I had change my code.redis cluster base on redis.UniversalClient, pls check!

zhuzihao-hz:
Hi👋 , I tried UniversalClient and UniversalClient, but got the following error while running

```text
[ERROR] PluginDaemonInternalServerError: redis: Watch requires at least one key
goroutine 1575 [running]:
```

Is this related to this func ``client.Watch(ctx, func(tx *redis.Tx)`` ?

Sorry, I'm not very familiar with the Redis API.

chenzikun:
> @zhuzihao-hz @chenzikun 使用redis集群，本地启动dify debug，会向plugin发送请求：response = self._request_with_plugin_daemon_response("POST", f"plugin/{tenant_id}/debugging/key", Response)，然后plugin会报错：[ERROR]PluginDaemonInternalServerError: redis: Watch require at least one key
> 
> 定位到redis.go文件Transaction方法里面的报错位置为： return client.Watch(ctx, func(tx *redis.Tx) error { 原因为： client.Watch(ctx, fn) // 缺少keys参数，导致Redis报错。
> 
> 高版本的redis，比如：go-redis v9.5.x ，Watch()方法源码中仍然保留以下判断逻辑 func (c *Client) Watch(ctx context.Context, fn func(*Tx) error,keys ...string) error { if len(keys) == 0 { return error.New("redis: Watch需要至少一个key") } ... }

fixed this error!

chenzikun:
@Yeuoly pls review this feature

PR #271, opened 2025-05-08
chore: dockerfile improvements

bowenliang123:
- dockerfile improvements
  - correct `FROM ... as` to `FROM ... AS` for setting base image, eliminating the warning message in docker image build `WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)`
  - move the copying action of `main` to latter steps , to reuse the previous built layer as many as possible especially for repeated builds and also help to extraction layer compressions earlier for image pulling
  - squeeze the file copying into single command to reduce image layers

### DIFF
--- a/docker/local.dockerfile
+++ b/docker/local.dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.23-alpine as builder
+FROM golang:1.23-alpine AS builder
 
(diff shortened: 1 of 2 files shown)


Yeuoly (approved):
LGTM

PR #278, opened 2025-05-10
dep: bump dify_plugin in plugin dependency template to 0.2.x

bowenliang123:
- bump dify_plugin in plugin dependency template to 0.2.x


bowenliang123:
OK. Let me change the code.

PR #281, opened 2025-05-12
fix:fix gorm id specified twice when use mysql,make field.HasDefaultV…

rouxiaomin:
When use mysql,Column 'id' specified twice when generate the sql

Failed to insert into MySQL, referred to this Gorm issue `https://github.com/go-gorm/gorm/issues/5893` and fix the error.
image


rouxiaomin:
fixed already in main branch

Issue #296, opened 2025-05-22
init eIf environment for plugin failed, the task is still updated as successful


NeatGuyCoding:
Running the dify-plugin-daemon without the required packages, such as "uv," may result in the installation task appearing successful, even though it has actually failed.

log message:
```
2025/05/22 02:57:35 full_duplex.go:27: [INFO]new plugin logged in: langgenius/bedrock:0.0.19
2025/05/22 02:57:35 full_duplex.go:59: [INFO]init environment for plugin langgenius/bedrock:0.0.19
2025/05/22 02:57:35 full_duplex.go:65: [ERROR]init environment failed: failed to find uv path: exit status 1, retrying
2025/05/22 02:57:35 full_duplex.go:59: [INFO]init environment for plugin langgenius/bedrock:0.0.19
2025/05/22 02:57:35 full_duplex.go:65: [ERROR]init environment failed: failed to find uv path: exit status 1, retrying
2025/05/22 02:57:35 full_duplex.go:59: [INFO]init environment for plugin langgenius/bedrock:0.0.19
2025/05/22 02:57:35 full_duplex.go:65: [ERROR]init environment failed: failed to find uv path: exit status 1, retrying
2025/05/22 02:57:35 full_duplex.go:59: [INFO]init environment for plugin langgenius/bedrock:0.0.19
2025/05/22 02:57:35 full_duplex.go:65: [ERROR]init environment failed: failed to find uv path: exit status 1, retrying
2025/05/22 02:57:35 full_duplex.go:29: [INFO]plugin langgenius/bedrock:0.0.19 has exited
```
dify ui:

![Image](https://github.com/user-attachments/assets/43754a68-6317-4376-a5ef-e4b278b741c9)

from database:

![Image](https://github.com/user-attachments/assets/d58b66ee-4725-43d8-95fb-8f8a24e00bb9)

PR #297, opened 2025-05-22
fix: errChan failed to write response because of panic nil (#296)

NeatGuyCoding:


Yeuoly (changes requested):
Can we join `err` and `er` into a single error using `errors.Join`? It will makes error message more user-friendly.

NeatGuyCoding:
> Can we join `err` and `er` into a single error using `errors.Join`? It will makes error message more user-friendly.

Modified with thanks

NeatGuyCoding (commented):
Fixed

Issue #299, opened 2025-05-22
[FEATURE] Add Request and Response Log for debugging purposes


NeatGuyCoding:
To aid in debugging, it's crucial to maintain request and response logs that include a trace ID, whether sourced from OTLP or generated internally.

For example:
```
2025/05/22 15:39:40 http_server.go:276: [DEBUG]HTTP Request(a57a712f-eb16-457b-ab18-92dbf5263787): POST /plugin/9b271264-72e6-4af9-ab96-8a94d802a574/debugging/key, content-type: 
2025/05/22 15:39:40 http_server.go:292: [DEBUG]HTTP Response(a57a712f-eb16-457b-ab18-92dbf5263787) status: 200, content-type: application/json; charset=utf-8, time elapsed: 1.009208ms, body: {"code":0,"message":"success","data":{"key":"e20acd40-c97d-4e26-a3a5-3dce7be5d89b"}}
```

PR #300, opened 2025-05-22
feature: add request and response log with request id(#299)

NeatGuyCoding:


Yeuoly:
Maybe what you want has already been implemented through `sentry`, checkout `SENTRY_ENABLED` and you can it

NeatGuyCoding:
> Maybe what you want has already been implemented through `sentry`, checkout `SENTRY_ENABLED` and you can it

Yeah, but it would need sentry and the body is not captured

I think it would be better for debug to show these info without sentry

Yeuoly:
I have several different thoughts on this:
1. Monitoring is a complex and difficult system project on production environment, we have both `sentry` `opentelemetry`, they are all designed to do this, you may want lots of information like `status_code` `headers` `machine info` `sqls` `p95/p99` and others, rather than implement it ourselves, using mature is better, as you know, implement it in a simple way is too rough.
2. If you want to see detailed request tracing on debugging environment, you need to change the default logger of Gin, instead of adding a new one, and at most cases, `breakpoint` could meet your needs.
3. Disabling this logger by default is a better choice, otherwise, logs are going to chaos.

Yeuoly:
I agree, that's why I said we need to change default logger of Gin if more tracing informations are required, but disabling this feature by default.

Issue #309, opened 2025-05-24
if tool response body is too long, will return empty


sunli1223:
Recently, while using MCP, the process often gets interrupted when the tool returns content that is too large. After debugging, it was found that both LLM and the tool's outputs are normal, but the agent fails to receive the output.

There are similar feedbacks in the issue list,
[1 https://github.com/langgenius/dify/issues/18501](https://github.com/langgenius/dify/issues/18501)
[2 https://github.com/langgenius/dify/issues/18501](https://github.com/langgenius/dify/issues/18731)
[3 https://github.com/langgenius/dify/issues/18501](https://github.com/langgenius/dify/issues/19978)
 but they have not been resolved. After several days of debugging, I have basically identified the cause of the problem.
the bug is in "internal/utils/http_requests/http_warpper.go",  Modifying as shown in the image can fix the issue. However, I hope the official team can include this in the configuration file.

<img width="2069" alt="Image" src="https://github.com/user-attachments/assets/60e56ca2-298c-4d08-9f73-a3cefa73a89b" />

PR #310, opened 2025-05-25
#fix llm response is too long,the scanner.Scan can't read

sunli1223:
this is for issue https://github.com/langgenius/dify-plugin-daemon/issues/309

### DIFF
--- a/internal/utils/http_requests/http_warpper.go
+++ b/internal/utils/http_requests/http_warpper.go
@@ -75,6 +75,44 @@ func PatchAndParse[T any](client *http.Client, url string, options ...HttpOption
 	return RequestAndParse[T](client, url, "PATCH", options...)
 }
 
+func readBodyStream(resp io.ReadCloser, callback func(data []byte) error) error {
+	reader := bufio.NewReaderSize(resp, 4*1024) // init with 4KB buffer
+	defer resp.Close()
+	for {
+		// read line by line
+		data, err := reader.ReadBytes('\n')
+		if err != nil {
+			if err != io.EOF {
+				log.Error("read body err:", err)
+			}
+			break
+		}
+
+		data = bytes.TrimSpace(data)
+		if len(data) == 0 {
+			continue
+		}
+
+		if bytes.HasPrefix(data, []byte("data:")) {
+			// split
+			data = data[5:]
+		}
+
+		if bytes.HasPrefix(data, []byte("event:")) {
+			// TODO: handle event
+			continue
+		}
+
+		// trim space
+		data = bytes.TrimSpace(data)
+		if err := callback(data); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 func RequestAndParseStream[T any](client *http.Client, url string, method string, options ...HttpOptions) (*stream.Stream[T], error) {
 	resp, err := Request(client, url, method, options...)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/utils/http_requests/http_warpper.go
```diff
@@ -75,6 +75,44 @@ func PatchAndParse[T any](client *http.Client, url string, options ...HttpOption
 	return RequestAndParse[T](client, url, "PATCH", options...)
 }
 
+func readBodyStream(resp io.ReadCloser, callback func(data []byte) error) error {
+	reader := bufio.NewReaderSize(resp, 4*1024) // init with 4KB buffer
```
Yeuoly, line 79: actually, the default buffer max size of bufio.Scanner is `64*1024`, it's big enough, I'm not sure what's your specific scenarios, but `4*1024` should not works.

sunli1223: @Yeuoly 
64k is not enough, as many current LLMs support more than 64k of context. If MCP is used, it will add all the prompt words, the MCP response, and all the bytes of the JSON format structure, making it easy to exceed this limit.  
```
reader := bufio.NewReaderSize(resp, 4*1024) // init with 4KB buffer
```
The 4K here is just a buffer to improve performance, and there is no maximum buffer limit.Therefore, the maximum byte size of a line ultimately depends on the context size supported by the LLM and the size of the response returned by the tool.

Yeuoly: In fact, Dify uses stream mode for all the LLM calls, response was split into chunks,  each chunk should not have a size larger than 64k.

As for the unlimited buffer, I suppose it's not a good design, it leads to DoS attack. for example, I can create a fake OpenAI server and returns 1G data in a single chunk, that's terrible, it might be a configurable environment variable, but not hardcoded. anyway, the buffer should not to be too large, if you use Dify in personal scenarios, yeah, just make it configurable.

sunli1223: @Yeuoly 
in my case, the llm return a chunk as below，when use tool calling, every chunk include the prompt_messages, the prompt_messages is very big:
```json
{
    "data": {
        "model": "deepseek-v3-250324",
        "prompt_messages": [
            {
                "role": "system",
                "content": "system promt(Text in Chinese with a length of 9197 characters.)",
                "name": ""
            },
            {
                "role": "user",
                "content": "user  query",
                "name": ""
            },
            {
                "role": "assistant",
                "content": "",
                "name": ""
            },
            {
                "role": "tool",
                "content": "Tool execution result: [{'type': 'text', 'text': 'too response in Chinese with a length of 29008 characters",
                "name": "hotel_info"
            }
        ],
        "system_fingerprint": "",
        "delta": {
            "index": 1,
            "message": {
                "role": "assistant",
                "content": "###",
                "name": "",
                "tool_calls": []
            },
            "usage": null,
            "finish_reason": null
        }
    },
    "error": ""
}
```

In Go, a Chinese character generally occupies 3 bytes, so although the text doesn't seem too long, multiplying it by 3 could surpass 64K.

@Yeuoly

tushverma:
Hey @Nov1c444 - I tried the change in my system - https://github.com/langgenius/dify/pull/20391 , but was still getting the same issue. In my case I am pulling data from my database which has long strings in one of the columns. Those strings are of character length > 70k. 
I added @sunli1223 's change and it worked as expected.

PR #323, opened 2025-06-01
fix build error, go.mod upgrade for github.com/panjf2000/ants/v2

zhanluxianshen:
image
I want to build and test it locally. But i found some errors.
The [panjf2000-ants](https://github.com/panjf2000/ants/tags?after=v2.0.0-rc.1) has deleted  v1.3.0.
So we should upgrade for match the right version.
image


Yeuoly (approved):
LGTM

Issue #331, opened 2025-06-07
damon panic when set ollama model


alwqx:
**Self Checks**

To make sure we get to you in time, please check the following :)
- [x] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify-plugin-daemon/issues), including closed ones.
- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [X] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [X] "Please do not modify this template :) and fill in all the required fields."

**Versions**
1. dify-plugin-daemon Version
2. dify-api Version

**Describe the bug**

when set ollama LLM model, `plugin_damon` panic. log below:
```shell
[GIN] 2025/06/07 - 13:17:44 | 200 |     417.208µs |      172.24.0.9 | GET      "/plugin/3cf018b4-64bf-4151-9d72-2419f8dfd022/management/install/tasks?page=1&page_size=100"
[GIN] 2025/06/07 - 13:17:49 | 200 |     374.299µs |      172.24.0.9 | GET      "/plugin/3cf018b4-64bf-4151-9d72-2419f8dfd022/management/install/tasks?page=1&page_size=100"
[GIN] 2025/06/07 - 13:17:54 | 200 |     386.488µs |      172.24.0.9 | GET      "/plugin/3cf018b4-64bf-4151-9d72-2419f8dfd022/management/install/tasks?page=1&page_size=100"
[GIN] 2025/06/07 - 13:17:59 | 200 |     436.088µs |      172.24.0.9 | GET      "/plugin/3cf018b4-64bf-4151-9d72-2419f8dfd022/management/install/tasks?page=1&page_size=100"
[GIN] 2025/06/07 - 13:18:01 | 200 |     454.218µs |      172.24.0.9 | GET      "/plugin/3cf018b4-64bf-4151-9d72-2419f8dfd022/management/models?page=1&page_size=256"

[GIN] 2025/06/07 - 13:18:01 | 500 |    1.807572ms |      172.24.0.9 | POST     "/plugin/3cf018b4-64bf-4151-9d72-2419f8dfd022/dispatch/model/validate_model_credentials"

2025/06/07 13:18:01 [Recovery] 2025/06/07 - 13:18:01 panic recovered:
runtime error: invalid memory address or nil pointer dereference
[...]

PR #332, opened 2025-06-08
fix: install numpy for ollama to prevent panic #331

alwqx:


Yeuoly:
You need to add numpy as a dependency here https://github.com/langgenius/dify-official-plugins/tree/main/models/ollama

PR #340, opened 2025-06-10
add USE_AWS_S3 args avoid ambiguity.

41tair:
Previously, we used the **S3_USE_AWS_MANAGED_IAM** environment variable, passing it to a **UseAws** argument to determine whether to use the AWS S3 API. This approach was confusing.

Now, we have introduced a new variable, **USE_AWS_S3**, to control this behavior. The original **S3_USE_AWS_MANAGED_IAM** variable will now be used for its intended purpose: to determine how to authorize S3 API requests.


Yeuoly (approved):
LGTM

PR #345, opened 2025-06-13
refactor(local_runtime): optimize listener lookup in stdioHolder

guanz42:
Directly fetch and invoke the listener for a given session_id instead of iterating over the entire listener map.


Yeuoly (approved):
LGTM

PR #364, opened 2025-06-26
feat:make response buffer size configurable

kinoooolu:
> [!IMPORTANT]
>
> 1. Make sure you have read our [contribution guidelines](https://github.com/langgenius/dify/blob/main/CONTRIBUTING.md)
> 2. Ensure there is an associated issue and you have been assigned to it
> 3. Use the correct syntax to link this PR: `Fixes #<issue number>`.

## Summary
Make the response buffer size of invoking tools/llm/agent strategy configurable instead of hard code a magic number.

## Checklist
- [ ] This change requires a documentation update
- [x] I understand that this PR may be closed in case there was no previous discussion or issues. (This doesn't apply to typos!)
- [ ] I've added a test for each change that was introduced, and I tried as much as possible to make a single atomic change.
- [ ] I've updated the documentation accordingly.
- [x] I have tested the new feature in local


Yeuoly:
Hi, controllers and services are generated automatically by `//go:generate go run ../../cmd/codegen/main.go`, you need to change its definitions here `internal/server/controllers/definitions.go`

kinoooolu:
After fix the buffer overflow in PR #362, making the buffer size configurable became less important. Additionally, this will involve modifying code templates, so discard it.

Issue #123, opened 2025-03-20
is there any example config for s3 storgae ?


system-man:

DavideDelbianco:
I use it with self hosted MinIO S3 storage
All I had to do after the release of v0.0.6-local was to add those ENV variables to the plugin_daemon service in docker-compose:

PLUGIN_STORAGE_TYPE:  "aws_s3"
PLUGIN_STORAGE_OSS_BUCKET:  "dify-plugin"
S3_USE_AWS_MANAGED_IAM: false
S3_ENDPOINT: minio:9000
AWS_ACCESS_KEY: your-minio-access-key
AWS_SECRET_KEY: your-minio-secret-key
AWS_REGION: auto

steveGuRen:
> I use it with self hosted MinIO S3 storage All I had to do after the release of v0.0.6-local was to add those ENV variables to the plugin_daemon service in docker-compose:
> 
> PLUGIN_STORAGE_TYPE: "aws_s3" PLUGIN_STORAGE_OSS_BUCKET: "dify-plugin" S3_USE_AWS_MANAGED_IAM: false S3_ENDPOINT: minio:9000 AWS_ACCESS_KEY: your-minio-access-key AWS_SECRET_KEY: your-minio-secret-key AWS_REGION: auto

hi, is this work?

DavideDelbianco:
Why would I reply with instructions on what to change?
It's written in English, either try it or read the commits of this repo...

PR #396, opened 2025-07-11
fix(service/controller): add support for paginated and non-paginated plugin list responses with backward compatibility

homejim:
… plugin list responses with backward compatibility

## Description

Fixes https://github.com/langgenius/dify/issues/22250
During the upgrade process, we encountered an error (reference issue: xxx). I resolved this by introducing a new response_type parameter in the Plugin Daemon.

**Fix Details:**

1. Added an optional response_type parameter to the /plugin/{tenantId}/management/list API:
 -  If response_type=paged is specified, the endpoint returns a paginated response.
 -  Otherwise, it returns the legacy non-paginated format, ensuring backward compatibility with older versions of Dify.

2. In Dify v1.6.0, the response_type=paged parameter is now included by default when calling this endpoint, aligning it with the updated Plugin Daemon behavior.

**Recommended Upgrade Paths:**

- **For users who have not yet upgraded:**

It's recommended to upgrade Plugin Daemon first, followed by Dify. This ensures a smooth and error-free transition.

- **For users who have already upgraded:**

You should first upgrade Dify to v1.6.0 before upgrading Plugin Daemon, to ensure compatibility between components.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

## Additional Information

### DIFF
--- a/internal/server/controllers/plugins.go
+++ b/internal/server/controllers/plugins.go
@@ -241,11 +241,12 @@ func FetchPluginFromIdentifier(c *gin.Context) {
 
 func ListPlugins(c *gin.Context) {
 	BindRequest(c, func(request struct {
-		TenantID string `uri:"tenant_id" validate:"required"`
-		Page     int    `form:"page" validate:"required,min=1"`
-		PageSize int    `form:"page_size" validate:"required,min=1,max=256"`
+		TenantID     string `uri:"tenant_id" validate:"required"`
+		Page         int    `form:"page" validate:"required,min=1"`
+		PageSize     int    `form:"page_size" validate:"required,min=1,max=256"`
+		ResponseType string `form:"response_type" validate:"omitempty,oneof=paged"`
 	}) {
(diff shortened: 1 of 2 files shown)


homejim:
@Yeuoly @crazywoola

Yeuoly:
As for me, it's not a good implementation that returns data with different schema, it's not so restful, and of course, versions between Dify and daemon should always align, upgrading only one side would make lots of problems, not only this one, such as MCP and Tools OAuth

homejim:
> As for me, it's not a good implementation that returns data with different schema, it's not so restful, and of course, versions between Dify and daemon should always align, upgrading only one side would make lots of problems, not only this one, such as MCP and Tools OAuth

You're right about this issue. This is just a solution provided based on the current state, mainly to solve the problems we encountered during the upgrade process. It's essential that all component versions remain consistent. However, during the upgrade process, if dify and daemon are deployed separately, this situation will cause problems now. @Yeuoly

PR #440, opened 2025-09-02
feat: support orphan plugin

wylswz:
## Description

- Allows installing plugin runtime without associating it with any tenant.
- Reinstall now generates tasks just like standard installations, so that the logs are visible from management console.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

### DIFF
--- a/internal/service/install_plugin.go
+++ b/internal/service/install_plugin.go
@@ -18,6 +18,7 @@ import (
 	"github.com/langgenius/dify-plugin-daemon/internal/utils/routine"
 	"github.com/langgenius/dify-plugin-daemon/internal/utils/stream"
 	"github.com/langgenius/dify-plugin-daemon/pkg/entities"
+	"github.com/langgenius/dify-plugin-daemon/pkg/entities/constants"
 	"github.com/langgenius/dify-plugin-daemon/pkg/entities/plugin_entities"
 	"github.com/langgenius/dify-plugin-daemon/pkg/plugin_packager/decoder"
 	"gorm.io/gorm"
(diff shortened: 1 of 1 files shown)


Yeuoly:
This implementation may lead an unexpected situation, under the community edition internal/service/install_plugin.goL690, plugin runtime will be removed if the refers was decreased to 0, its behavior is different from enterprise edition.

wylswz:
> This implementation may lead an unexpected situation, under the community edition internal/service/install_plugin.goL690, plugin runtime will be removed if the refers was decreased to 0, its behavior is different from enterprise edition.

I handled that in `atomic.go`, however, leaving refers as it is seems to be a more consistent approach, I think this field is invisible to community & SaaS users.

Yeuoly:
> > This implementation may lead an unexpected situation, under the community edition internal/service/install_plugin.goL690, plugin runtime will be removed if the refers was decreased to 0, its behavior is different from enterprise edition.
> 
> I handled that in `atomic.go`, however, leaving refers as it is seems to be a more consistent approach, I think this field is invisible to community & SaaS users.

What I mean is that if the "refers" count equals 0, it's an important flag indicating that the plugin has been completely uninstalled and is no longer needed by any tenant. In this case, the plugin manager will shut down the plugin's subprocess (for the community edition). 

When this plugin is installed again, it will fail because the record still exists in the database, and there is no subprocess to launch. Consequently, a new plugin runtime will not be initiated.

wylswz:
> > > This implementation may lead an unexpected situation, under the community edition internal/service/install_plugin.goL690, plugin runtime will be removed if the refers was decreased to 0, its behavior is different from enterprise edition.
> > 
> > 
> > I handled that in `atomic.go`, however, leaving refers as it is seems to be a more consistent approach, I think this field is invisible to community & SaaS users.
> 
> What I mean is that if the "refers" count equals 0, it's an important flag indicating that the plugin has been completely uninstalled and is no longer needed by any tenant. In this case, the plugin manager will shut down the plugin's subprocess (for the community edition).
> 
> When this plugin is installed again, it will fail because the record still exists in the database, and there is no subprocess to launch. Consequently, a new plugin runtime will not be initiated.

Makes sense, I'll try not to modify the behavior of reference counting

wylswz:
I've modified the implementation

- The semantic of `refers` is now consistent.
- The daemon automatically adds a reference from "global tenant" when the plugin is installed.

Yeuoly:
I  need to confirm whether this approach is to use `global_tenant_installation` to prevent refers equaling to 0?

internal/server/controllers/plugins.go
```diff
@@ -121,6 +122,10 @@ func InstallPluginFromIdentifiers(app *app.Config) gin.HandlerFunc {
 			if request.Metas == nil {
 				request.Metas = []map[string]any{}
 			}
+			if request.TenantID == constants.GlobalTenantId && !app.PluginAllowOrphans {
```
Yeuoly, line 129: This check is required in `Upgrade` also.

wylswz:
> I need to confirm whether this approach is to use `global_tenant_installation` to prevent refers equaling to 0?

Yes.

internal/server/controllers/plugins.go
wylswz: Done.

wylswz:
Added another commit that logs the reinstallation

PR #487, opened 2025-10-31
feat: support passthrough propagation to plugins

ywqwancbx:
## Passthrough Propagation to Plugins

- Add top-level `passthrough` to `InvokePluginRequest` and `Session`, and include it in `Session.Message()` so it reaches the Python SDK as a first-class field (similar to `conversation_id`).
- Keep `InvokeTool` data payload strictly for business parameters; carry `passthrough` via session to avoid schema drift.
- Remove temporary debug logs and unused helpers; retain only necessary code for passthrough propagation.
- Backward compatible; existing tools continue to work without changes.
- Corresponding upstream changes in Dify main repo: [feat: add passthrough parameter support for workflow tools](https://github.com/langgenius/dify/pull/27078).

### DIFF
--- a/internal/core/session_manager/session.go
+++ b/internal/core/session_manager/session.go
@@ -40,6 +40,7 @@ type Session struct {
 	AppID          *string        `json:"app_id"`
 	EndpointID     *string        `json:"endpoint_id"`
 	Context        map[string]any `json:"context"`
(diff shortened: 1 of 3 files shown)


Yeuoly:
Close as no plan

PR #499, opened 2025-11-07
refactor: plugin lifecycle control panel

Yeuoly:
## Description

Please provide a brief description of the changes made in this pull request.
Please also include the issue number if this is related to an issue using the format `Fixes #123` or `Closes #123`.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing
- [ ] I have tested the changes locally and confirmed they work as expected
- [ ] I have added unit tests where necessary and they pass successfully

### Bug Fix (if applicable)
- [ ] I have used GitHub syntax to close the related issue (e.g., `Fixes #123` or `Closes #123`)

## Additional Information

Please provide any additional context that would help reviewers understand the changes.

PR #525, opened 2025-12-08
fix: fix panic error

fatelei:
## Description

fix https://github.com/langgenius/dify/issues/29282

The nil pointer dereference happens when:
1. A session is created via NewSession() where runtime field remains nil
2. The session is used before BindRuntime() is called to attach a runtime instance
3. GenericInvokePlugin tries to access the runtime's Listen() method

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

## Additional Information


Yeuoly:
This issue is basically because of a bad initialization process, which daemon passed through a request to runtime while  the runtime was not ready to accept it, it should already been fixed in #499 which is not been released yet.

PR #548, opened 2025-12-22
chore: init listener first avoid panic error

fatelei:
## Description

fix #29959, init listener first to avoid panic error

```go
func(sessionId string, data []byte) {
    // FIX: avoid deadlock to plugin invoke
    s.l.Lock()
    listener := s.listener[sessionId]
    s.l.Unlock()
    if listener != nil {
        listener(data)
    }
}
```

if the listener is nil, it will panic, so init listener first

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

## Additional Information

### DIFF
--- a/internal/core/local_runtime/instance.go
+++ b/internal/core/local_runtime/instance.go
@@ -71,9 +71,9 @@ func newPluginInstance(
 		errReader:              errReader,
 		l:                      &sync.Mutex{},
 		appConfig:              appConfig,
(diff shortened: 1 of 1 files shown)


Yeuoly (approved):
LGTM

PR #552, opened 2025-12-25
Feat/Implement structured logging and Trace ID propagation

### LINKED ISSUE #551: Implement structured logging and Trace ID propagation
**Self Checks**

- [X] I have searched for existing issues [search for existing issues](https://github.com/langgenius/dify-plugin-daemon/issues), including closed ones.
- [X] I confirm that I am using English to submit this report (我已阅读并同意 [Language Policy](https://github.com/langgenius/dify/issues/1542)).
- [X] [FOR CHINESE USERS] 请务必使用英文提交 Issue，否则会被关闭。谢谢！:)
- [X] "Please do not modify this template :) and fill in all the required fields."

**Is your feature request related to a problem? Please describe.**
1. Logs are unstructured, making them difficult to be parsed by modern logging systems.
2. Missing Trace ID and Span ID in the context.
3. Requests returning 404 status codes are not logged.

**Describe the solution you'd like**
1. Adopt a unified structured logging format
2. Extract and propagate Trace ID and Span ID throughout the application context.
3. Inject Trace ID into headers when making outbound HTTP requests.
4. Ensure backward compatibility (e.g., via configuration flags) to avoid breaking existing log ingestion pipelines.

**Describe alternatives you've considered**

**Additional context**

41tair:
## Description
Implement structured logging and Trace ID propagation
Closes #551 
## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing
- [X] I have tested the changes locally and confirmed they work as expected
- [X] I have added unit tests where necessary and they pass successfully

### Bug Fix (if applicable)
- [X] I have used GitHub syntax to close the related issue (e.g., `Fixes #123` or `Closes #123`)

## Additional Information
image

### DIFF
--- a/internal/core/dify_invocation/calldify/types.go
+++ b/internal/core/dify_invocation/calldify/types.go
@@ -12,18 +12,18 @@ type RealBackwardsInvocation struct {
 	client              *http.Client
 	writeTimeout        int64
 	readTimeout         int64
-	ctx                 context.Context
+	traceCtx            context.Context
 }
 
 func (i *RealBackwardsInvocation) SetContext(ctx context.Context) {
-	i.ctx = ctx
+	i.traceCtx = ctx
 }
 
 func (i *RealBackwardsInvocation) Context() context.Context {
-	if i.ctx == nil {
+	if i.traceCtx == nil {
 		return context.Background()
 	}
-	return i.ctx
+	return i.traceCtx
 }
 
 type BaseBackwardsInvocationResponse[T any] struct {


Stream29:
@41tair I have a question: Will the information in the `context`, like `traceId`, be lost with `log.Info(...)`?

I have this question because there's 2 variants of `log.Info`:
- `func Info(msg string, args ...any)`
- `func InfoContext(ctx context.Context, msg string, args ...any)`

The first variant just uses `context.Background()` as `context`.

In my view, we can remove the first variant to prevent possible misusage.
It's always better to pass the `ctx` param. If there is no `context.Context`, the caller should explicitly pass `context.Background()` as `context`.

41tair:
> @41tair I have a question: Will the information in the `context`, like `traceId`, be lost with `log.Info(...)`?
> 
> I have this question because there's 2 variants of `log.Info`:
> 
> * `func Info(msg string, args ...any)`
> * `func InfoContext(ctx context.Context, msg string, args ...any)`
> 
> The first variant just uses `context.Background()` as `context`.
> 
> In my view, we can remove the first variant to prevent possible misusage. It's always better to pass the `ctx` param. If there is no `context.Context`, the caller should explicitly pass `context.Background()` as `context`.

The daemon supports multiple commands beyond just web requests. In many scenarios (such as CLI), passing context is unnecessary; mandating context.Background() for every call would introduce significant boilerplate.

Stream29 (approved):

Stream29:
Then it looks good to me.

Yeuoly:
Since daemon may work in a distributed environment, backwards invocation replies on `session_id` stored in Redis to trace the identity from user.

At the very beginning of each request to plugin, daemon generates a unique transaction id and waits for callbacks from plugin, but under the serverless mode, we cannot guarantee the session context is retrievable in memory, so, I just simply marshal `TraceContext` and `Identity` into cache, then use them in 

```go
func (h *ServerlessTransactionHandler) Handle(ctx *gin.Context, sessionId string)
```

 cc @41tair

Stream29:
@Yeuoly Thank you for your excellent work! I didn't consider these issues in the distributed scenarios so clearly. I should learn from this.

PR #564, opened 2026-01-10
feat: add redis distribute lock

### LINKED ISSUE #541: install plugin in cluster env, InitPythonEnvironment will delete .venv dir causing race conditions and failures
**Self Checks**

**Versions**
1. dify-plugin-daemon Version 0.5.1
2. dify-api Version 1.10.1-fix.1

**Describe the bug**
When running multiple pods of dify-plugin-daemon in k8s, the `InitPythonEnvironment()` function can be executed concurrently by different pods for the same plugin. This race condition may cause the `deleteVirtualEnvironment()` function to be triggered while another pod is still initializing the Python virtual environment, resulting in the virtual environment being deleted during the initialization process. This leads to initialization failures and unpredictable behavior.

**Root Cause**
The issue is triggered by the `handleNewLocalPlugins()` function in the cluster mode:

1. Each pod runs `startLocalMonitor()` which calls `handleNewLocalPlugins()` every 30 seconds
2. `handleNewLocalPlugins()` lists all installed plugins from the shared `installedBucket` (cross-pod shared storage)
3. Each pod checks if the plugin is already running locally (`c.localPluginRuntimes.Exists()`)
4. If not running locally, the pod calls `LaunchLocalPlugin()` -> `InitEnvironment()` -> `InitPythonEnvironment()`
5. Although `LaunchLocalPlugin()` has a `localPluginInstallationLock`, this lock is in-memory and only prevents concurrent launches within the same pod
6. Multiple pods can simultaneously pass this local lock check and all attempt to initialize the same plugin's Python environment in the shared storage
7. This causes race conditions where one pod's `deleteVirtualEnvironment()` destroys the environment being built by another pod

**To Reproduce**
1. Deploy dify-plugin-daemon with multiple replicas (e.g., 3+ pods) in a Kubernetes cluster
2. Share the same persistent volume or storage for plugin working directories across all pods
3. Install or update a plugin that triggers Python environment initialization (it take about 1 minute to initialize)
4. Observe that multiple pods attempt to initialize the Python environment simultaneously
5. See error: virtual environment gets deleted while being initialized, causing "virtual environment is invalid" errors

**Expected behavior**
- Only one pod should be allowed to initialize the Python environment for a specific plugin at a time
- Other pods should wait for the initialization to complete or timeout gracefully
- The initialization process should be protected by a distributed lock mechanism
- Log messages should be emitted when the virtual environment is being deleted and recreated

fatelei:
## Description

fix #541 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

## Additional Information

### DIFF
--- a/internal/core/control_panel/launcher_local.go
+++ b/internal/core/control_panel/launcher_local.go
@@ -71,7 +71,7 @@ func (c *ControlPanel) LaunchLocalPlugin(
 		// expire: generous upper bound for env initialization; tryLockTimeout: wait up to the same duration
 		expire := 15 * time.Minute
 		tryTimeout := 2 * time.Minute
-		log.Info("acquiring distributed init lock, plugin", pluginUniqueIdentifier.String(), "expire", expire.String())
+		log.Info("acquiring distributed init lock", "plugin", pluginUniqueIdentifier.String(), "expire", expire.String())
 		if err := cache.Lock(lockKey, expire, tryTimeout); err != nil {
 			// failed to acquire the lock within timeout
 			err = errors.Join(err, fmt.Errorf("failed to acquire distributed env-init lock"))
@@ -84,9 +84,9 @@ func (c *ControlPanel) LaunchLocalPlugin(
 		}
 		defer func() {
 			if unlockErr := cache.Unlock(lockKey); unlockErr != nil {
-				log.Warn("failed to release distributed init lock", pluginUniqueIdentifier.String(), "error", unlockErr.Error())
+				log.Warn("failed to release distributed init lock", "plugin", pluginUniqueIdentifier.String(), "error", unlockErr.Error())
 			} else {
-				log.Info("released distributed init lock", pluginUniqueIdentifier.String())
+				log.Info("released distributed init lock", "plugin", pluginUniqueIdentifier.String())
 			}
 		}()
 
--- a/internal/core/local_runtime/setup_python_environment.go
+++ b/internal/core/local_runtime/setup_python_environment.go
@@ -284,7 +284,7 @@ func (p *LocalPluginRuntime) deleteVirtualEnvironment() error {
 		}
 		return err
 	}
-	log.Warn("deleting existing Python virtual environment for %s at %s", p.Config.Identity(), venvDir)
+	log.Warn("deleting existing Python virtual environment", "plugin", p.Config.Identity(), "path", venvDir)
 	return os.RemoveAll(venvDir)
 }
 


internal/core/control_panel/launcher_local.go
```diff
@@ -62,15 +65,41 @@ func (c *ControlPanel) LaunchLocalPlugin(
...
-		// release semaphore
-		releaseLockAndSemaphore()
-		return nil, nil, err
+	// To avoid cross-pod races on Python venv creation, guard InitEnvironment with a Redis-based distributed lock.
+	{
+		lockKey := fmt.Sprintf("env_init_lock:%s", pluginUniqueIdentifier.String())
+		// expire: generous upper bound for env initialization; tryLockTimeout: wait up to the same duration
+		expire := 15 * time.Minute
+		tryTimeout := 2 * time.Minute
+		log.Info("acquiring distributed init lock for %s (expire=%s)...", pluginUniqueIdentifier.String(), expire.String())
```
Yeuoly: Use structured log instead.

internal/core/local_runtime/setup_python_environment.go
```diff
@@ -225,11 +225,12 @@ func (p *LocalPluginRuntime) checkPythonVirtualEnvironment() (*PythonVirtualEnvi
 
 func (p *LocalPluginRuntime) deleteVirtualEnvironment() error {
 	// check if virtual environment exists
-	if _, err := os.Stat(path.Join(p.State.WorkingPath, envPath)); err != nil {
+	venvDir := path.Join(p.State.WorkingPath, envPath)
+	if _, err := os.Stat(venvDir); err != nil {
```
Yeuoly, line 281: Should check specific error type like os.ErrNotFound

pkg/utils/cache/redis_lock_test.go
```diff
@@ -0,0 +1,68 @@
+package cache
+
+import (
+	"testing"
+	"time"
+
+	miniredis "github.com/alicebob/miniredis/v2"
+)
+
+func TestDistributedLockExclusive(t *testing.T) {
```
Yeuoly: The tests already exist.

internal/core/local_runtime/setup_python_environment.go
fatelei: if dir is not exists, i think we can success, for idempotent

internal/core/control_panel/launcher_local.go
fatelei: log.Info / log.Warn use 

```
func logWithCaller(ctx context.Context, level slog.Level, msg string, args ...any) {
	logger := slog.Default()
	if !logger.Enabled(ctx, level) {
		return
	}
	var pc uintptr
	var pcs [1]uintptr
	runtime.Callers(3, pcs[:])
	pc = pcs[0]
	r := slog.NewRecord(time.Now(), level, msg, pc)
	r.Add(args...)
	_ = logger.Handler().Handle(ctx, r)
}
```

it is already structure log

Yeuoly: I mean you may use 
```go
log.Info("acquiring distributed init lock, plugin", pluginUniqueIdentifier.String(), "expire", expire.String())
```

fatelei: update

Yeuoly (approved):

PR #589, opened 2026-01-28
fix: synchronize remote debugging plugins across cluster nodes

### LINKED ISSUE #588: remote debug return no node available
**Self Checks**

**Versions**
1. dify-plugin-daemon Version
2. dify-api 1.11.4

**Describe the bug**
remote debug return no node available

Image

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

fatelei:
fix #588 #590 https://github.com/langgenius/dify/issues/31684

## Problem
Remote debugging plugins were not being synchronized across cluster nodes, causing "no plugin available nodes found" errors when trying to invoke plugins from different nodes.

## Root Causes
1. **Remote debugging plugins not registered to cluster** - The `ClusterTunnel` notifier was not being added to ControlPanel
2. **Plugin ID inconsistency** - Remote plugins used different plugin_id formats during installation vs. querying
3. **Non-idempotent registration** - `RegisterPlugin` failed on reconnection with "plugin has been registered" error

### Core Fixes
- **internal/types/models/curd/atomic.go**:
  - Unify plugin_id calculation for remote plugins (author/name without version)
  - Remove plugin_id from plugin query conditions
  - Clear old cache when plugin_id is updated

- **internal/cluster/plugin.go**:
  - Make `RegisterPlugin` idempotent by updating existing plugin instead of returning error

- **internal/core/control_panel/daemon.go**:
  - Add cluster field to ControlPanel
  - Add SetCluster() method for lazy cluster initialization

- **internal/core/control_panel/server_debugger.go**:
  - Register remote debugging plugins to cluster on connection
  - Unregister from cluster on disconnection

- **internal/core/plugin_manager/manager.go**:
  - Add SetCluster() method to set cluster after initialization

- **internal/server/server.go**:
  - Call SetCluster() instead of AddClusterTunnel()

### Design Decision
Only remote debugging plugins are synchronized across cluster nodes. Local plugins run only on the node where they are installed and are not registered to the cluster.

### Additional Improvements
- Error handling improvements using `errors.Is()` instead of `==`
- Handle 404 for missing plugin assets gracefully
- Handle already-installed debugging plugins gracefully

## Testing
- Remote debugging plugin can be invoked from any node in the cluster
- Plugin reconnection works without errors
- Cache invalidation works correctly when plugin_id changes

## Description

## Type of Change

- [x] Bug fix
- [x] New feature
- [x] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

## Additional Information

### DIFF
--- a/internal/cluster/plugin.go
+++ b/internal/cluster/plugin.go
@@ -37,7 +37,19 @@ func (c *Cluster) RegisterPlugin(lifetime plugin_entities.PluginLifetime) error 
 	}
 
 	if c.plugins.Exists(identity.String()) {
-		return errors.New("plugin has been registered")
+		// idempotent: plugin already registered, just update its state
+		if existing, ok := c.plugins.Load(identity.String()); ok {
+			// update the lifetime reference
+			existing.lifetime = lifetime
+			// update plugin state immediately
+			err = c.doPluginStateUpdate(existing)
+			if err != nil {
+				return errors.Join(err, errors.New("failed to update plugin state"))
+			}
+		} else {
+			log.Warn("plugin does not exist", "identity", identity.String())
+		}
+		return nil
 	}
 
 	l := &pluginLifeTime{
@@ -49,7 +61,7 @@ func (c *Cluster) RegisterPlugin(lifetime plugin_entities.PluginLifetime) error 
 	// do plugin state update immediately
 	err = c.doPluginStateUpdate(l)
 	if err != nil {
-return errors.Join(err, errors.New("failed to update plugin state"))
+		return errors.Join(err, errors.New("failed to update plugin state"))
 	}
 
 	if c.showLog {
@@ -143,10 +155,6 @@ func (c *Cluster) doPluginStateUpdate(lifetime *pluginLifeTime) error {
 		return err
 	}
 
-	if c.showLog {
-		log.Info("updating plugin state", "identity", identity.String())
-	}
-
 	hashedIdentity := plugin_entities.HashedIdentity(identity.String())
 
 	scheduleState := &pluginState{
@@ -167,9 +175,6 @@ func (c *Cluster) doPluginStateUpdate(lifetime *pluginLifeTime) error {
 			return err
 		}
 	} else {
-		if c.showLog {
-			log.Info("updating plugin state", "identity", identity.String())
-		}
 		// update plugin state
 		scheduleState.ScheduledAt = &[]time.Time{time.Now()}[0]
 		err = cache.SetMapOneField(PLUGIN_STATE_MAP_KEY, stateKey, scheduleState)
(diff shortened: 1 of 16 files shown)


internal/server/controllers/plugins.go
```diff
@@ -18,6 +18,9 @@ func GetAsset(c *gin.Context) {
 	asset, err := pluginManager.GetAsset(c.Param("id"))
 
 	if err != nil {
+		if strings.Contains(err.Error(), "no such file or directory") {
```
Stream29, line 21: I wonder is it good to ignore error by it's error message. Hard-coding such strings is a bad practice.

internal/cluster/plugin.go
```diff
@@ -37,7 +37,17 @@ func (c *Cluster) RegisterPlugin(lifetime plugin_entities.PluginLifetime) error
...
+		if existing, ok := c.plugins.Load(identity.String()); ok {
+			// update the lifetime reference
+			existing.lifetime = lifetime
+			// update plugin state immediately
+			err = c.doPluginStateUpdate(existing)
+			if err != nil {
+				return errors.Join(err, errors.New("failed to update plugin state"))
+			}
+		}
+		return nil
```
Stream29, line 52: At least, we should do something when not ok.

internal/cluster/plugin_test.go
```diff
@@ -140,6 +140,107 @@ func TestPluginScheduleLifetime(t *testing.T) {
...
+		return
+	}
+
+	if len(nodes) != 1 {
+		t.Errorf("plugin not available after re-registration")
+		return
+	}
+}
+
 // TODO: I need to implement this test, now it's randomly working
```
Stream29: ??? Randomly working means it's not fixed.

internal/server/controllers/plugins.go
```diff
@@ -18,6 +18,9 @@ func GetAsset(c *gin.Context) {
...
+			return
```
Stream29, line 22: We shouldn't return without giving any response to the request.

internal/types/models/curd/atomic.go
```diff
@@ -68,12 +79,26 @@ func InstallPlugin(
 		} else if err != nil {
 			return err
 		} else {
+			// Update plugin_id if it differs (e.g., for remote plugins)
+			oldPluginID := p.PluginID
+			if p.PluginID != pluginID {
+				log.Info("Updating plugin_id", "from", p.PluginID, "to", pluginID)
+				p.PluginID = pluginID
+			}
```
Stream29, line 87: These behaviors should be tested: original bahavior with normal plugins should keep the same and debugging plugins should be installed correctly, having correct database records, even with debugging plugin reconnect.

internal/types/models/curd/atomic.go
```diff
@@ -244,28 +281,23 @@ func UninstallPlugin(
...
+			return installation.PluginID
 		}
 
 		// delete tool installation
 		if declaration.Tool != nil {
 			toolInstallation := &models.ToolInstallation{
-				PluginID: pluginToBeReturns.PluginID,
+				PluginID: getPluginID(),
 				TenantID: tenantId,
 			}
```
Stream29, line 302: Similarly, these behaviors should be tested: correct both for normal plugins and debugging plugins

Stream29 (changes requested):

internal/server/controllers/plugins.go
fatelei: because when no asset it will failed

Stream29:
You can re-request a review if you think it's ready.

PR #608, opened 2026-02-13
fix: remove internal .env loading from plugin daemon

weichun-wang:
## Summary

Remove internal `.env` loading from PluginDaemon.

PluginDaemon now relies only on process environment variables provided by the invoker/runtime layer.

## Why

This fixes the startup issue in containerized environments where configuration is injected directly and no physical `.env` file exists.

It also aligns with the principle that configuration injection should be handled by the invoker/runtime layer rather than by PluginDaemon itself.

## Changes

- remove internal `.env` loading from `cmd/server/main.go`
- rely only on process environment variables during startup


QuantumGhost:
I prefer removing the `.env` loading logic (removing the invocation of `godotenv.Load()`)  inside PluginDaemon and let the invoker of PluginDaemon decides how to inject configuration information.

This PR seems to introduce more configuration complexity than it reduces.

As for debugging, I suggest utilizing [the godotenv command line tool](https://github.com/joho/godotenv) and injecting configuration with the following commnad:

```bash
godotenv -f .env dify-plugin ...
```

weichun-wang:
Thanks for the feedback — that makes sense.
I agree that PluginDaemon itself does not need to handle .env loading, and configuration injection should be decided by the invoker/runtime layer.


My original goal was to fix the startup failure when no physical .env file exists, but your suggestion is cleaner and better aligned with containerized deployment.


I’ll update the PR accordingly and remove the internal .env loading logic.






大河剧
***@***.***

PR #625, opened 2026-03-05
fix: upgrade plugin fails when new version declaration is not yet in DB

euxx:
## Description

When `UpgradePlugin()` is called for auto-upgrades, it unconditionally called `CombinedGetPluginDeclaration(newPluginUniqueIdentifier)` before checking whether the new version package had been downloaded. Since `models.PluginDeclaration` only contains records for already-installed packages, this lookup returned `ErrPluginNotFound` for any version not yet present — causing an immediate error **before the async download task was ever started**.

**Fix**: check `models.Plugin` existence first.
- New version already in DB → fetch declaration and upgrade synchronously (existing behaviour).
- New version not yet downloaded → start download with `NewDeclaration: nil`; inside `ProcessUpgradeJob`, fetch declaration after `PluginInstallEventDone` fires (plugin is now installed, record exists) and complete the upgrade.

## Type of Change

- [x] Bug fix

### Testing

### Bug Fix (if applicable)

## Additional Information

The daemon returns HTTP 200 for all plugin operations; errors are encoded in the response body. Callers (Dify API workers) therefore received a 200 response containing an error payload, making this failure silent in logs unless the body was inspected.

### DIFF
--- a/internal/tasks/install_plugin.go
+++ b/internal/tasks/install_plugin.go
@@ -121,15 +121,16 @@ func ProcessUpgradeJob(
 			// Fetch the new declaration from DB now that the package has been installed.
 			// It may have been unavailable (nil) at job creation time when the new version
 			// was not yet downloaded.
-			newDeclaration := job.NewDeclaration
-			if newDeclaration == nil {
-				decl, err := helper.CombinedGetPluginDeclaration(job.NewIdentifier, runtimeType)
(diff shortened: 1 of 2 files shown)


fatelei:
please add ut

euxx:
> please add ut

Added it. Please take a look.

PR #684, opened 2026-04-12
fix(persistence): correctly calculate storage size delta during save operations

leslie2046:
## Description

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

## Additional Information

📝 Background
Previously, the Save method in the persistence layer incorrectly updated the tenant's total storage size by always incrementing it with the full length of the new data. This resulted in over-reporting storage usage whenever an existing key was updated, leading to premature quota exhaustion.

🛠️ Changes
This PR refines the storage accounting logic to use a "delta-based" approach:

Existing Size Retrieval: Before saving, the system now checks if the key already exists and retrieves its current size.
Delta Calculation: It calculates the actual size difference (sizeDiff = new_size - old_size).
Optimized Update:
Storage quota checks are now performed against the sizeDiff rather than the total allocatedSize.
The database's aggregate Size field is updated using this calculated difference, ensuring accurately tracked usage for both new entries and overwrites.
✅ Impact
Accuracy: Tenant storage statistics will now accurately reflect the net change in data volume.
Reliability: Prevents "storage leaks" where the recorded size continually grows even if files are being replaced with smaller versions.

### DIFF
--- a/internal/core/persistence/persistence.go
+++ b/internal/core/persistence/persistence.go
@@ -47,18 +47,26 @@ func (c *Persistence) Save(tenantId string, pluginId string, maxSize int64, key 
 		maxSize = c.maxStorageSize
 	}
 
+	oldSize := int64(0)
+	if exists, err := c.storage.Exists(tenantId, pluginId, key); err == nil && exists {
+		if size, err := c.storage.StateSize(tenantId, pluginId, key); err == nil {
+			oldSize = size
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


leslie2046:
@fatelei cc

fatelei (approved):

PR #712, opened 2026-04-26
chore(ci): move image builds to depot

goocarlos:
Moves the daemon image workflow from local Docker builds on the runner to Depot remote container builds.

What changed:
- added `depot.json` and set the workflow to use `DEPOT_PROJECT_ID`
- switched the build job to `depot/setup-action@v1` + `depot/build-push-action@v1`
- added `permissions.id-token: write` for OIDC auth
- removed runner-local `docker build` / `docker tag` / `docker push` loops
- switched to digest artifacts plus a manifest-creation step for the final multi-arch tags

Cache strategy:
- removed GitHub Actions layer cache export/import
- rely on Depot's native layer cache instead of `type=gha`

Local developer scripts were not changed.

### DIFF
--- a/.github/workflows/build-push.yml
+++ b/.github/workflows/build-push.yml
@@ -43,8 +43,11 @@ jobs:
 
   build:
     needs: matrix_prepare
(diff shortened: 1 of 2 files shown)


fatelei:
rebase main code

PR #716, opened 2026-04-28
feat: add subcommand extract to slim

41tair:
## Description

Add the extract as sub command for daemon, used for extract plugin schema.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing
- [X] I have tested the changes locally and confirmed they work as expected
- [X] I have added unit tests where necessary and they pass successfully

### Bug Fix (if applicable)

## Additional Information

### DIFF
--- a/pkg/slim/extract.go
+++ b/pkg/slim/extract.go
@@ -1,0 +1,70 @@
+package slim
+
+import (
+	"encoding/json"
+	"io"
+
+	"github.com/langgenius/dify-plugin-daemon/pkg/entities/plugin_entities"
+	"github.com/langgenius/dify-plugin-daemon/pkg/plugin_packager/decoder"
+)
+
+const OutputJSON = "json"
+
+type ExtractOptions struct {
+	PluginID string
+	Action   string
+	Path     string
+	Output   string
+}
+
+type ExtractResult struct {
+	UniqueIdentifier string                            `json:"unique_identifier,omitempty"`
+	Manifest         plugin_entities.PluginDeclaration `json:"manifest"`
+	Verification     *decoder.Verification             `json:"verification,omitempty"`
+}
+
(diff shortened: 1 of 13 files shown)


pkg/slim/extract.go
```diff
@@ -0,0 +1,71 @@
...
+	TenantID string `json:"tenant_id,omitempty"`
+	UserID   string `json:"user_id,omitempty"`
+	Data     any    `json:"data"`
+}
+
+func RunExtract(cfg *SlimConfig, opts ExtractOptions, w io.Writer) error {
+	if opts.Output == "" {
+		opts.Output = OutputJSON
+	}
+	if opts.Output != OutputJSON {
```
fatelei, line 37: it seem useless ?

pkg/slim/extract.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	if opts.Action != "" {
+		payload, err := BuildActionArgs(result.Manifest, opts.Action)
+		if err != nil {
+			return err
+		}
+		args.Data = payload
+	}
+
+	if err := json.NewEncoder(w).Encode(args); err != nil {
```
fatelei, line 67: after encode, it seem not output

pkg/slim/extract.go
41tair: This is to prevent issues from occurring when the output consists of other parameters.Currently, we only support JSON.

pkg/slim/extract.go
41tair: The w in the encoder is an io.Writer.In the main function call, an os.Stdout will be passed in.

fatelei (approved):

PR #746, opened 2026-05-23
fix: support lowercase proxy environment variables as fallback

EvanYao826:
## Summary

The plugin_daemon config only reads uppercase `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` via envconfig tags. When only lowercase `http_proxy`, `https_proxy`, and `no_proxy` are set (common in many environments), the proxy settings are silently ignored, causing plugin installation failures behind HTTP proxies.

Add fallback in `SetDefault()` to check lowercase env vars when uppercase are not set, matching the behavior of other Dify containers (api, web).

## Fix

In `internal/types/app/default.go`, use the existing `setDefaultString` helper function for proxy environment variable fallback to maintain code consistency.

## Checklist

- [x] I have read the [Contributing Guide](https://github.com/langgenius/dify-plugin-daemon/blob/main/CONTRIBUTING.md)
- [x] Fixes #18752

### DIFF
--- a/internal/types/app/default.go
+++ b/internal/types/app/default.go
@@ -9,15 +9,9 @@ import (
 
 func (config *Config) SetDefault() {
 	// Fallback to lowercase proxy env vars if uppercase are not set
-	if config.HttpProxy == "" {
-		config.HttpProxy = os.Getenv("http_proxy")
-	}
-	if config.HttpsProxy == "" {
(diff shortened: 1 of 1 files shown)


EvanYao826:
Thank you for the review! I will update the PR to use the existing setDefaultString helper function for consistency.

fatelei:
main has support it

PR #750, opened 2026-05-29
feat: auto-detect pip mirror

BeautyyuYanli:
## Description

This PR adds startup-time automatic PyPI mirror selection for plugin dependency installation.

### Environment variables

- `PIP_MIRROR_AUTO_DETECT`
  - Default: `true`
  - Controls whether the daemon should try to auto-select a suitable PyPI mirror during startup.
  - When enabled and no explicit mirror is configured, the daemon detects the public IP region and may assign a local mirror for regions with poor network connectivity.

- `PIP_MIRROR_URL`
  - Default: empty
  - Manually specifies the PyPI mirror used for plugin dependency resolution.
  - When set, it takes precedence over auto detection and disables the auto-selected mirror.

### Detection behavior

- The daemon uses `https://cloudflare.com/cdn-cgi/trace` at startup to detect the public IP region.
- If the detected region needs a local mirror, `config.PipMirrorUrl` is populated before plugin runtimes install Python dependencies.
- Regions that do not need a mirror keep the default behavior.
- Detection failures are non-fatal: the daemon logs a warning and continues startup without changing the mirror configuration.

### Documentation and tests

- Updates `.env.example` and `README.md` to describe the mirror-related configuration behavior.
- Adds tests for config defaults, env overrides, trace parsing, mirror precedence, region-based application, and non-blocking failure handling.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

## Additional Information

Local verification performed:

```shell
go build -o /tmp/opencode/dify-plugin-daemon-server cmd/server/main.go
go test ./cmd/server ./internal/types/app
```

**This PR is drafted by `gpt-5.4` and `gpt-5.5`. I'm responsible for all the changes. I have reviewed the code and varified the behavior, while breaks may still exist. Reach me to fix in this case.**


cmd/server/main.go
```diff
@@ -10,6 +16,13 @@ import (
 	"github.com/langgenius/dify-plugin-daemon/pkg/utils/log"
 )
 
+const (
+	alibabaCloudPypiMirrorURL  = "https://mirrors.aliyun.com/pypi/simple/"
```
wylswz: I would suggest moving mirror candidate to a environment variable instead of a const.

Having multiple default candidates may help improve the reliability.

BeautyyuYanli: Those who want to set ENV may set the mirror directly? This AUTO DETECT is meant to provide a default and out-of-box value.

wylswz: I'm rethinking about the design. I'll try
- Not to introduce another dependency on CF. Instead, rely on the connectivity to the configured mirror.
- Make the connectivity issue visible to the user. If the mirrors are unavailable to the service (auto detected or explicitly configured, users should be able to see that installing plugins)

## Approach 1
1. Add a background checker, and expose the connectivity to healthiness probe of this service.
2. Add on indicator to Dify console.

## Approach 2
Do a connectivity check when starting up, fail fast with a clear log.

fatelei (approved):

PR #754, opened 2026-06-06
fix: support lowercase proxy environment variables

jcj46548-ux:
## Summary
- read lowercase `http_proxy`, `https_proxy`, and `no_proxy` as fallbacks when uppercase proxy config is not set
- pass both uppercase and lowercase proxy variables to local plugin processes and dependency installation commands
- add tests covering lowercase fallbacks, uppercase precedence, and generated proxy environment entries

## Tests
- `go test ./internal/types/app`
- `go test ./internal/core/local_runtime -run '^$'`

Full `go test ./internal/core/local_runtime` currently requires `uv`; the integration tests fail locally with `UV not found`.

Related issue: https://github.com/langgenius/dify/issues/18752


fatelei:
main has support it

PR #755, opened 2026-06-06
fix: prevent path traversal during plugin extraction

jcj46548-ux:
## Summary
- validate zip entry paths before extracting plugin packages
- reject parent-directory and backslash paths so entries cannot escape the target working directory
- preserve normal nested file extraction, including packages with explicit directory entries

## Tests
- `go test ./pkg/plugin_packager/decoder ./pkg/plugin_packager/packager`

Notes:
- `go test ./pkg/plugin_packager` is blocked locally because `pkg/license/private_key/PRIVATE_KEY.pem` is not present in the checkout.
- `go test -race ./pkg/plugin_packager/decoder` is blocked locally because the current Go environment has cgo disabled.

### DIFF
--- a/pkg/plugin_packager/decoder/zip.go
+++ b/pkg/plugin_packager/decoder/zip.go
@@ -18,6 +18,8 @@ import (
 	"github.com/langgenius/dify-plugin-daemon/pkg/utils/parser"
 )
 
+var errUnsafeZipPath = errors.New("unsafe path in plugin package")
+
 type ZipPluginDecoder struct {
 	PluginDecoder
 	PluginDecoderHelper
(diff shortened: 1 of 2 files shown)


pkg/plugin_packager/decoder/zip.go
```diff
@@ -326,6 +359,58 @@ func (z *ZipPluginDecoder) ExtractTo(dst string) error {
 	return nil
 }
 
+func safeExtractPath(dst, entryName string) (string, error) {
```
fatelei:
```
 func safeExtractPath(dst, entryName string) (string, error) {
  	if entryName == "" || strings.Contains(entryName, `\`) {
  		return "", fmt.Errorf("%w: %q", errUnsafeZipPath, entryName)
  	}

  	for _, part := range strings.Split(entryName, "/") {
  		if part == ".." {
  			return "", fmt.Errorf("%w: %q", errUnsafeZipPath, entryName)
  		}
  	}

  	entryPath := path.Clean(entryName)
  	if entryPath == "." || path.IsAbs(entryPath) {
  		return "", fmt.Errorf("%w: %q", errUnsafeZipPath, entryName)
  	}

  	targetPath := filepath.Join(dst, filepath.FromSlash(entryPath))
  	if !pathIsInside(dst, targetPath) {
  		return "", fmt.Errorf("%w: %q", errUnsafeZipPath, entryName)
  	}

  	return targetPath, nil
  }
```

pkg/plugin_packager/decoder/zip.go
```diff
@@ -297,27 +300,57 @@ func (z *ZipPluginDecoder) UniqueIdentity() (plugin_entities.PluginUniqueIdentif
...
-
-		bytes, err := z.ReadFile(filepath.Join(dir, filename))
-		if err != nil {
-			return err
-		}
-
-		filename = filepath.Join(workingPath, filename)
+	if z.reader == nil {
+		return z.err
+	}
```
fatelei, line 307:
```
root, err := os.OpenRoot(dst)
  	if err != nil {
  		return err
  	}
  	defer root.Close()

  	for _, file := range z.reader.File {
  		entryPath, err := safeEntryPath(file.Name)
  		if err != nil {
  			return err
  		}

  		if file.FileInfo().IsDir() {
  			if err := root.MkdirAll(entryPath, 0755); err != nil {
  				return err
  			}
  			continue
  		}

  		if err := root.MkdirAll(path.Dir(entryPath), 0755); err != nil {
  			return err
  		}

  		if err := extractZipFile(root, entryPath, file); err != nil {
  			return err
  		}
  	}

  	return nil
```

pkg/plugin_packager/decoder/zip.go
```diff
@@ -326,6 +350,64 @@ func (z *ZipPluginDecoder) ExtractTo(dst string) error {
...
+	for _, part := range strings.Split(entryName, "/") {
+		if part == ".." {
+			return "", fmt.Errorf("%w: %q", errUnsafeZipPath, entryName)
+		}
+	}
+
+	entryPath := path.Clean(entryName)
+	if entryPath == "." || path.IsAbs(entryPath) {
+		return "", fmt.Errorf("%w: %q", errUnsafeZipPath, entryName)
+	}
```
fatelei, line 367: not check targetPath := filepath.Join(dst, filepath.FromSlash(entryPath))
  	if !pathIsInside(dst, targetPath) {
  		return "", fmt.Errorf("%w: %q", errUnsafeZipPath, entryName)
  	} ?

fatelei:
another mr has been merged

PR #766, opened 2026-07-02
fix(local-runtime): recover stopped plugin instances

### LINKED ISSUE #765: Local runtime can keep routing requests to stopped plugin instances
**Self Checks**

**Versions**

1. dify-plugin-daemon Version: 0.6.3
2. dify-api Version: 1.4.2

**Describe the bug**

Local plugin requests can keep being routed to a plugin instance whose stdio has already been closed.

When this happens, dispatch fails before the request reaches the plugin process. A typical error looks like:

```text
PluginDaemonInternalServerError: write |1: file already closed
failed to write request
```

`|1` is the write side of the pipe used by plugin-daemon to send requests to the local plugin subprocess. If plugin-daemon closes that pipe through `PluginInstance.Stop()` or after stdout exits, the instance can still remain selectable in the local runtime's instance pool until asynchronous shutdown cleanup removes it.

For local plugins running with a single instance, that stopped instance can be the only candidate in the pool, so later requests may continue to select the same dead instance and fail immediately with `file already closed` or `broken pipe`.

**To Reproduce**

One way to reproduce the behavior is to hit the local runtime while a plugin instance has been stopped but has not yet been removed from the instance pool:

1. Run plugin-daemon in local runtime mode with a local plugin that has one running instance.
2. Trigger a condition that stops the plugin instance or closes its stdio, for example a heartbeat timeout or runtime stop path.
3. Before the asynchronous shutdown cleanup removes the instance from the runtime pool, send another dispatch request to the same local plugin.
4. Observe that the request can still be routed to the stopped instance.
5. See an error similar to:

```text
PluginDaemonInternalServerError: write |1: file already closed
failed to write request
```

**Expected behavior**

Once a local plugin instance is known to be stopped, or once its stdin write fails because the pipe is closed, plugin-daemon should:

- stop routing new requests to that instance
- evict it from the local runtime instance pool
- wake the scheduler so a replacement instance can be launched
- retry write-before-dispatch failures briefly when recovery is possible

The runtime should recover without requiring users to reinstall the plugin or restart plugin-daemon.

**Screenshots**

Not applicable.

**Additional context**

The root cause is that the local runtime relies on asynchronous shutdown cleanup to remove instances from the pool. There is no synchronous stopped flag on `PluginInstance`, and the load balancer/write path can still use an instance after its stdio has already been closed.

The scheduler also counts entries in the instance slice, so a stopped-but-not-yet-pruned instance can delay replacement.

This issue is about local runtime recovery after closed stdio/write-before-dispatch failures. It is not about plugin dependency installation failures or provider-specific validation logic.

Kevin9703:
## Description

Fixes #765

This pull request fixes a local runtime recovery issue where plugin-daemon can continue routing requests to a local plugin instance after that instance's stdio pipes have already been closed.

When this happens, dispatch fails before the request reaches the plugin process, for example:

```text
PluginDaemonInternalServerError: write |1: file already closed
failed to write request
```

The root cause is that stopped local plugin instances are removed from the runtime instance pool asynchronously. During that gap, the load balancer can still select the stale instance, and the write path can still attempt to write to its closed stdin pipe.

This PR makes the local runtime recover from that state by:

- marking `PluginInstance` as stopped synchronously when stdio is closed or stdout exits
- skipping stopped instances during local runtime load balancing
- evicting dead instances when stdin writes fail with closed-pipe style errors
- waking the scheduler after eviction so replacement instances can be started promptly
- retrying local write-before-dispatch failures within a small bounded window
- serializing writes to a plugin instance's stdin to avoid interleaved payloads from concurrent sessions

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Performance improvement
- [ ] Other

### Testing

### Bug Fix (if applicable)

## Additional Information

Local validation:

- `go test ./internal/core/local_runtime/... ./internal/core/io_tunnel/...`
- `go build ./cmd/server`

The change is limited to local plugin runtime instance lifecycle, load balancing, stdin write handling, and write-before-dispatch retry behavior. It does not change serverless runtime behavior.

### DIFF
--- a/internal/core/local_runtime/instance.go
+++ b/internal/core/local_runtime/instance.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os/exec"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/google/uuid"
@@ -33,6 +34,8 @@ type PluginInstance struct {
 
 	started  bool // mark the instance as started, it will be set to true when the first heartbeat is received
 	shutdown bool // mark the instance as shutdown, it will be set to true when stdout reader is closed
+	stopped  atomic.Bool
+	writeMu  sync.Mutex
 
 	// app config
 	appConfig *app.Config
@@ -105,6 +108,9 @@ func (s *PluginInstance) Error() error {
 
 // Stop stops the stdio, of course, it will shutdown the plugin asynchronously
 func (s *PluginInstance) Stop() {
+	if !s.stopped.CompareAndSwap(false, true) {
+		return
+	}
 	s.inWriter.Close()
 	s.outReader.Close()
 	s.errReader.Close()
@@ -125,6 +131,7 @@ func (s *PluginInstance) Stop() {
 // Once the subprocess exists itself, STDOUT always close, which results in `CLOSE STDOUT`
 func (s *PluginInstance) StartStdout() {
 	defer func() {
+		s.stopped.Store(true)
 		log.Info(
 			"plugin stdout reader exiting",
 			"plugin", s.pluginUniqueIdentifier,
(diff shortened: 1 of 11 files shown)


internal/core/local_runtime/instance.go
```diff
@@ -368,9 +375,26 @@ func (s *PluginInstance) Monitor() error {
 }
 
 func (s *PluginInstance) Write(data []byte) error {
+	if s.IsStopped() {
+		return ErrInstanceStopped
+	}
+
+	s.writeMu.Lock()
```
fatelei: i think this lock is useless

Kevin9703: You’re right, I’ll remove it.

fatelei (approved):

PR #780, opened 2026-07-27
feat(cache): cache the tenant model provider list in the daemon

GareArc:
## Description

Plugins assigned to a workspace from the enterprise admin console never appear in that workspace's model provider list (ESQ1-190, Zendesk #3547).

Dify caches the per-tenant model provider list in Redis, but enterprise installs plugins by calling this daemon directly, bypassing Dify's API. Dify therefore never learns its own cache is stale, and the newly assigned plugin stays invisible until the TTL expires. A cache is only correct if the system that owns the writes can invalidate it — so the cache moves here.

This is phase 3 of the plan. Phase 2 (langgenius/dify#39632) added flags to disable Dify's caches; this builds the daemon-side replacement so `PLUGIN_MODEL_PROVIDERS_CACHE_ENABLED` can be turned off.

### What changed

- `ListModels` now reads through `helper.CombinedListModelInstallations`, mirroring the existing `CombinedGetPluginDeclaration` / `DeletePluginDeclarationCache` pair in `combined.go`.
- Storage is a Redis hash per tenant, field `<page>:<pageSize>`, so one `DEL` drops every cached page — no `SCAN`, no version counter.
- `InstallPlugin` / `UninstallPlugin` / `UpgradePlugin` invalidate it, guarded on the same `declaration.Model != nil` condition that already guards the `AIModelInstallation` row writes.
- The invalidation is `defer`red rather than run after the transaction. A retried install aborts on `ErrPluginAlreadyInstalled` before reaching any post-transaction code, so a `DEL` placed there becomes unreachable once it has failed once.
- TTL defaults to 60 minutes, configurable via `PLUGIN_MODEL_INSTALLATIONS_CACHE_TTL`. It is a safety net behind explicit invalidation, not the primary mechanism.

### Separate first commit

`parser.UnmarshalJson` could never return a slice: `validator.Struct` rejects any non-struct top-level value, so it always errored before returning, which made `cache.GetMapField[[]T]` unusable. Slices now skip validation alongside maps and strings. `UnmarshalJsonBytes2Slice` remains the variant that validates per element.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [x] Performance improvement
- [ ] Other

### Testing

`pkg/utils/cache/helper` and `pkg/utils/parser` tests pass locally. The `internal/types/models/curd` tests require postgres and redis via `TestMain`, so they run in CI but were not executed locally.

## Additional Information

The model *schema* cache has the same ownership defect but a much narrower blast radius, and it is not what #3547 reports. It moves in a follow-up: #781.

Known and deliberate here:

- `cmd/commandline/plugin/cleanup.go` deletes `AIModelInstallation` rows with no invalidation. It initialises only the DB, never Redis, so it cannot invalidate at all — after a manual cleanup run, deleted providers linger for up to the TTL. Consistent with how that CLI already bypasses every other daemon cache.
- Read-repopulate race: a reader holding pre-commit rows can write them back after the `DEL`. Bounded by the TTL. Same exposure as the existing declaration cache.
- Inert in production until the Dify-side flag is turned off.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -235,6 +235,11 @@ PLUGIN_RUNTIME_MAX_BUFFER_SIZE=5242880
 # plugin install/upgrade timeout in minutes
 PLUGIN_INSTALL_TIMEOUT=15
 
+# cache the tenant's model installation list in redis. Turn it on only where dify runs with
+# PLUGIN_MODEL_PROVIDERS_CACHE_ENABLED=false, otherwise both sides hold the same payload
+# and the redis footprint doubles.
+PLUGIN_MODEL_INSTALLATIONS_CACHE_ENABLED=false
+
 # how long a tenant's cached model installation list survives without an explicit invalidation, in minutes
 PLUGIN_MODEL_INSTALLATIONS_CACHE_TTL=1440
 
--- a/.env.example
+++ b/.env.example
@@ -236,7 +236,7 @@ PLUGIN_RUNTIME_MAX_BUFFER_SIZE=5242880
 PLUGIN_INSTALL_TIMEOUT=15
 
 # how long a tenant's cached model installation list survives without an explicit invalidation, in minutes
-PLUGIN_MODEL_INSTALLATIONS_CACHE_TTL=60
+PLUGIN_MODEL_INSTALLATIONS_CACHE_TTL=1440
 
 # dify backwards invocation write timeout in milliseconds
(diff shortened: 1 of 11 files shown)


wylswz:
I suggest disabling this cache by default, otherwise the cache size is doubled, i.e., another scaling up needs to be scheduled before deploying to cloud production env.

wylswz (approved):

PR #796, opened 2026-08-07
fix(packager): reject zip entries that escape the extraction directory

SashaMIT:
## Description

`ZipPluginDecoder.ExtractTo` builds the output path with `path.Join(dst, dir)` / `filepath.Join(workingPath, filename)` from raw zip entry names. `path.Join` cleans the result but does not contain it, so an entry named `../../evil.py` resolves **outside** the plugin working directory.

Impact: installing an attacker-crafted `.difypkg` yields arbitrary file write on the daemon host - overwriting runtime files, or planting code that the daemon subsequently loads/executes - anywhere signature verification is disabled (local/self-hosted installs) or the package otherwise reaches extraction.

## Type of Change

- [x] Bug fix

### Testing

Each entry is now resolved against the destination with `filepath.Rel` and rejected unless strictly contained (`illegal file path in plugin package`). On rejection the partially-written working directory is removed, matching the existing error path.

Added `pkg/plugin_packager/decoder/zip_extract_test.go`:
- `TestExtractToRejectsPathTraversal`: a `../../` entry errors and nothing is written outside `dst`
- `TestExtractToAllowsNormalEntries`: nested legitimate entries still extract correctly

`go test ./pkg/plugin_packager/decoder/ -run TestExtractTo -count=1 -v` - both pass.

## Additional Information

The read paths (`ReadFile`/`ReadDir`) are unaffected; this hardens only the write sink. Behavior for well-formed packages (including marketplace packages) is unchanged.

### DIFF
--- a/pkg/plugin_packager/decoder/zip.go
+++ b/pkg/plugin_packager/decoder/zip.go
@@ -298,21 +298,29 @@ func (z *ZipPluginDecoder) UniqueIdentity() (plugin_entities.PluginUniqueIdentif
 func (z *ZipPluginDecoder) ExtractTo(dst string) error {
 	// copy to working directory
 	if err := z.Walk(func(filename, dir string) error {
-		workingPath := path.Join(dst, dir)
+		entryName := path.Join(dir, filename)
+		target := filepath.Join(dst, filepath.FromSlash(entryName))
+
+		// zip entry names are attacker-controlled; path.Join only cleans the
+		// result, so a "../" prefix escapes dst into an arbitrary write. Reject
+		// any entry whose resolved path is not contained in dst.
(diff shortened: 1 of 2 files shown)


fatelei:
you can use os.OpenRoot

SashaMIT:
Good call, done. Writes now go through `os.OpenRoot(dst)` (`root.MkdirAll` / `root.OpenFile`), so extraction is contained at the filesystem layer rather than only by the lexical check. Kept the `filepath.Rel` guard as well: it produces the clear "illegal file path in plugin package" error for traversal entries, while OpenRoot enforces containment even for anything the check misses. Module targets Go 1.26 so the API is available; decoder tests pass.

fatelei (approved):

PR #799, opened 2026-08-07
fix(local-runtime): stop inheriting daemon environment in plugin subprocesses

SashaMIT:
Plugin subprocesses no longer receive a copy of the daemon's own environment variables.

## What

`getInstanceCmd` built plugin processes with `cmd.Environ()`, which copies the daemon process's entire environment into every plugin subprocess. In every real deployment that environment carries the daemon's full credential set: `DB_PASSWORD` / `DB_USERNAME` (direct Postgres access to all tenant plugin data), `SERVER_KEY` (the shared key protecting the daemon's management and dispatch routes), `DIFY_INNER_API_KEY` (the daemon's identity for Dify's internal API), Redis and cloud storage credentials, `ADMIN_API_KEY`, and anything else the operator places in the container env. Any installed plugin could read all of it from `os.environ` and exfiltrate it over the network, since plugins make arbitrary outbound calls by design. The slim CLI's local mode had the same pattern (`os.Environ()`), running marketplace-downloaded plugin code with the developer's full shell environment.

## Why

The local runtime's design (separate process per plugin, dedicated working directory, stdio-only IPC, heartbeat watchdog) exists to run third-party marketplace code at reduced trust. Inheriting the daemon environment defeats the confidentiality half of that isolation, and plugins need none of these variables to function: daemon to plugin traffic is stdio-framed, and plugin configuration arrives per-invocation rather than through the environment. The repo already recognized this pattern for the short-lived `uv` installer child process (`buildUVCommandEnv`, with a test asserting `UNRELATED_SECRET` is not inherited); the long-lived plugin process was missed.

## How

Replace inheritance with an explicit allowlist builder, `BuildPluginCommandEnv`, mirroring `buildUVCommandEnv`. It passes through what plugins legitimately need: `PATH`, `HOME`, locale variables (`LANG`, `LC_ALL`, `LC_CTYPE`), temp directories (`TMPDIR`, `TEMP`, `TMP`), `TZ`, custom CA bundles (`SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`), and proxy variables in both cases, with the daemon config's proxy settings (`HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY`, themselves loaded from env or config) taking precedence. `INSTALL_METHOD=local` is set exactly as before. Everything else, including all `*_KEY` / `*_SECRET` / `*_PASSWORD` / `*_TOKEN` style credentials, never reaches plugin code. The slim CLI local mode now uses the same builder.

## How verified

- New `internal/core/local_runtime/subprocess_test.go`: `TestBuildPluginCommandEnv` asserts allowlisted variables pass through, config proxy settings win over inherited ones, and `DB_PASSWORD` / `SERVER_KEY` / `DIFY_INNER_API_KEY` / `AWS_*` / `REDIS_PASSWORD` / `ADMIN_API_KEY` are absent. `TestGetInstanceCmdDoesNotInheritDaemonEnv` builds the real plugin command against a fake venv and asserts the same on `cmd.Env`, proving the spawn site is wired to the allowlist.
- `go build` passes for the touched packages, and `go test ./internal/core/local_runtime/... ./pkg/slim/...` passes in full. (A whole-repo `go build ./...` additionally requires the gitignored `pkg/license/private_key/PRIVATE_KEY.pem`, unrelated to this change.)

Same disclosure class as our merged #796: a one-line omission visible in the public source, fixed directly via public PR.

### DIFF
--- a/internal/core/local_runtime/subprocess.go
+++ b/internal/core/local_runtime/subprocess.go
@@ -9,6 +9,7 @@ import (
 	"slices"
 	"time"
 
+	"github.com/langgenius/dify-plugin-daemon/internal/types/app"
 	"github.com/langgenius/dify-plugin-daemon/pkg/entities/constants"
 	"github.com/langgenius/dify-plugin-daemon/pkg/entities/plugin_entities"
 	routinepkg "github.com/langgenius/dify-plugin-daemon/pkg/routine"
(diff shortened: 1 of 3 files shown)


fatelei (approved):

SashaMIT:
Thanks for the review and the merge, @fatelei.

PR #809, opened 2026-08-31
fix(io_tunnel): kill local plugin subprocess on stream close

koc-Z3:
## Summary

When the upstream consumer closes the streaming response (e.g. Dify API
calls `invoke_result.close()` after the user clicks Stop), the daemon
correctly detects the disconnect via gin's `CloseNotify` and runs the
`Stream`'s `OnClose` callbacks. But the existing `closeListener`
callback only removes the session's listener from a map — it does NOT
stop the local plugin subprocess that holds the open
`requests.post(stream=True)` connection to the upstream LLM provider.

Result: the upstream provider keeps generating tokens until natural
end of generation. The Dify UI hides new output but the upstream
provider is still consuming GPU/quota. Bug has been reported across
multiple Dify issues for ~19 months with no upstream fix.

This PR fixes the cancel propagation by extending the existing
`OnClose` registration in `invokePluginOnce` to also call
`instance.Stop()` when the runtime is `*LocalPluginRuntime`.
`Stop()` is idempotent (uses `sync/atomic.CompareAndSwap`) and closes
the subprocess's stdin/stdout/stderr pipes, which causes the
subprocess to exit and its open HTTP connection to the upstream
provider to close within milliseconds.

## Root cause

The codebase comment at `internal/core/local_runtime/instance.go:117-125`
documents the relationship explicitly:

```
CLOSE STDOUT = KILL and REAP subprocess
CLOSE INSTANCE = CLOSE STDOUT
```

Closing the listener does NOT close the instance; closing the instance
is what kills the subprocess. None of the existing cancel paths
(`writer.CloseNotify()`, `context.AfterFunc(ctx, response.Close)`,
`OnClose(closeListener)`) close the instance.

The full cancel propagation chain is:

1. ✅ Dify API: `invoke_result.close()` → `GeneratorExit` propagates
   through the streaming generator chain → `httpx.Response.close()` →
   TCP closed to plugin_daemon.
2. ✅ plugin_daemon: gin's `ctx.Request.Context()` is cancelled →
   `baseSSEService.writer.CloseNotify()` fires →
   `pluginDaemonResponse.Close()`.
3. ✅ plugin_daemon: `context.AfterFunc(ctx, response.Close)` at
   `generic.go:89` also fires → `Stream.Close()` runs the registered
   `onClose` callbacks.
4. ❌ Before this PR: the `onClose` callback registered at
   `generic.go:191` as `closeListener` only removes the listener
   from a map (`instance.removeStdioHandlerListener(sessionId)` at
   `local_runtime/instance.go:92`). It does NOT call
   `instance.Stop()`.
5. ❌ The plugin subprocess (Python with an open
   `requests.post(stream=True)` to MTPLX/Ollama/OpenAI/etc.) is never
   killed. The subprocess keeps streaming until the upstream finishes
   its natural generation. MTPLX logs typically show `elapsed_s: 8.8s,
   completion_tokens: 500` even after a 1-second Dify stop.

After this PR, step 4 also calls `instance.Stop()` when the runtime
is `*LocalPluginRuntime`, which closes the subprocess's pipes and
cancels the upstream generation within milliseconds.

## Changes

  - **`internal/core/io_tunnel/generic.go`** — `invokePluginOnce`: the
    `OnClose` callback now type-asserts `runtime` to
    `*local_runtime.LocalPluginRuntime` and calls `instance.Stop()` on
    the local plugin instance for the session, before delegating to
    the existing `closeListener`. The old behavior is preserved for
    non-local runtimes (debugging, serverless, remote) which already
    have their own cleanup paths.
  - **`internal/core/local_runtime/type.go`** — added
    `LookupSession(sessionID string) (*PluginInstance, bool)`
    accessor on `LocalPluginRuntime` so external callers can find the
    instance bound to a session without poking at the unexported
    `sessionToInstanceMap`.
  - **`internal/core/io_tunnel/cancel_test.go`** — new test
    `TestInvokePluginOnceOnCloseFires` verifies that the cancel
    wiring is reachable on `Stream.Close()` for the non-local-runtime
    path. This is the regression boundary: when the runtime is local,
    the new code calls `instance.Stop()`; when it isn't, the old
    behavior is preserved.
  - **`internal/core/local_runtime/type_test.go`** — new tests
    `TestLookupSessionReturns{False,Instance}After{Store,Delete}`
    verify the `LookupSession` accessor contract.

## Tested

  - `go build ./internal/core/io_tunnel/... ./internal/core/local_runtime/...` — OK
  - `go vet ./...` — OK (only a pre-existing pattern-match warning for
    a missing test fixture file)
  - `go test -count=1 -short ./internal/core/io_tunnel/... ./internal/core/local_runtime/...` — OK
  - End-to-end (verified before this PR via direct MTPLX test):
    `requests.post(stream=True)` + `r.close()` after 1s → MTPLX log
    shows `cancellation_elapsed_s: 1.01`, 26 tokens. After this PR
    lands, the same test through Dify should match (currently 8.82s
    and 500 tokens).

## Related Dify issues

The same underlying bug is reported across multiple Dify issues:

  - langgenius/dify#36164 "Stop Response button does not immediately
    terminate LLM calls and keeps consuming tokens" (open, 1.13.3)
  - langgenius/dify#40966 "Stop generating not work" (open, 1.15.0)
  - langgenius/dify#6493 "LLM node cannot stopped with ollama" (open
    since 0.6.14)
  - langgenius/dify#13035 "stop response api ineffective for chat
    assistant" (closed as not planned, 0.15.2)
  - langgenius/dify#35549 "chatflows remain in running state in log
    system" (closed via #37129, partial log-only fix)

The companion Dify API-side patches are in PR #41520 in langgenius/dify
(chat-mode + agent-chat `is_stopped` checks). Without the daemon-side
fix in this PR, the API-side patches correctly end the Dify stream in
1-2s but MTPLX/upstream providers keep generating until natural
completion. With this PR, the cancel propagates end-to-end.

## Backwards compatibility

  - Non-local runtimes (debugging, serverless, remote) take the
    existing code path unchanged. Only `*LocalPluginRuntime` is
    affected.
  - `instance.Stop()` is idempotent — already-protected by
    `CompareAndSwap` inside `PluginInstance.Stop` at
    `internal/core/local_runtime/instance.go:109-116`. Safe to call
    multiple times or after the subprocess has already exited.
  - The `LookupSession` accessor is purely additive — does not
    change any existing public API.
  - If a session has been re-bound to a different instance between
    `OnClose` registration and `OnClose` firing, `IsStopped()` on the
    returned instance protects against stopping an instance that is
    currently serving another session.

### DIFF
--- a/internal/core/io_tunnel/generic.go
+++ b/internal/core/io_tunnel/generic.go
@@ -188,7 +188,26 @@ func invokePluginOnce[Req any, Rsp any](
 				pluginMap,
 			)
 			if err == nil {
-				response.OnClose(closeListener)
+				// On stream close, kill the local plugin subprocess too — the
+				// upstream listener (closeListener) only removes the listener
+				// from a map and does not stop the subprocess, so its open
+				// connection to the upstream LLM provider stays alive and the
+				// provider keeps generating tokens until natural completion.
+				// Closing the subprocess's stdin/stdout/stderr pipes via
+				// instance.Stop() (PluginInstance.Stop closes the pipes — see
+				// internal/core/local_runtime/instance.go:117-125) forces the
+				// subprocess to exit, which closes its requests.post(stream=True)
+				// connection and cancels the upstream generation. Idempotent —
+				// safe to call multiple times or after the subprocess has
+				// already exited.
+				response.OnClose(func() {
+					if localRuntime, ok := runtime.(*local_runtime.LocalPluginRuntime); ok {
+						if instance, ok := localRuntime.LookupSession(session.ID); ok && instance != nil && !instance.IsStopped() {
+							instance.Stop()
+						}
+					}
+					closeListener()
+				})
 				return response, nil
 			}
 
--- a/internal/core/io_tunnel/cancel_test.go
+++ b/internal/core/io_tunnel/cancel_test.go
@@ -1,0 +1,101 @@
+package io_tunnel
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/langgenius/dify-plugin-daemon/internal/core/io_tunnel/access_types"
+	"github.com/langgenius/dify-plugin-daemon/internal/core/session_manager"
+	"github.com/langgenius/dify-plugin-daemon/pkg/entities"
+	plugin_entities "github.com/langgenius/dify-plugin-daemon/pkg/entities/plugin_entities"
+	"github.com/stretchr/testify/require"
+)
+
+// fakeNonLocalRuntime implements PluginRuntimeSessionIOInterface but is NOT
+// a *local_runtime.LocalPluginRuntime, so invokePluginOnce takes the
+// existing onclose-only path (not the new instance.Stop() path). This is
+// the regression boundary: when the runtime is local, the new code calls
+// instance.Stop(); when it isn't, the old behavior is preserved.
+type fakeNonLocalRuntime struct{}
+
+func (r *fakeNonLocalRuntime) Type() plugin_entities.PluginRuntimeType {
+	return plugin_entities.PLUGIN_RUNTIME_TYPE_REMOTE
+}
+func (r *fakeNonLocalRuntime) Configuration() *plugin_entities.PluginDeclaration { return nil }
+func (r *fakeNonLocalRuntime) Identity() (plugin_entities.PluginUniqueIdentifier, error) {
+	return "test/plugin:0.0.1", nil
+}
+func (r *fakeNonLocalRuntime) HashedIdentity() (string, error) { return "hashed", nil }
+func (r *fakeNonLocalRuntime) Checksum() (string, error)      { return "checksum", nil }
+func (r *fakeNonLocalRuntime) Listen(sessionID string) (*entities.Broadcast[plugin_entities.SessionMessage], error) {
+	return entities.NewCallbackHandler[plugin_entities.SessionMessage](), nil
+}
+func (r *fakeNonLocalRuntime) Write(sessionID string, _ access_types.PluginAccessAction, _ []byte) error {
+	return context.Canceled
+}
+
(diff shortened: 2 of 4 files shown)


internal/core/io_tunnel/generic.go
```diff
@@ -188,7 +188,26 @@ func invokePluginOnce[Req any, Rsp any](
...
+				// provider keeps generating tokens until natural completion.
+				// Closing the subprocess's stdin/stdout/stderr pipes via
+				// instance.Stop() (PluginInstance.Stop closes the pipes — see
+				// internal/core/local_runtime/instance.go:117-125) forces the
+				// subprocess to exit, which closes its requests.post(stream=True)
+				// connection and cancels the upstream generation. Idempotent —
+				// safe to call multiple times or after the subprocess has
+				// already exited.
+				response.OnClose(func() {
+					if localRuntime, ok := runtime.(*local_runtime.LocalPluginRuntime); ok {
```
fatelei, line 204: This change fixes the missing cancellation path, but it is unsafe as implemented.

  Stream.Close() is also called on normal completion (SESSION_MESSAGE_TYPE_END), so this callback will stop the local plugin process after every successful invocation, not only after a client disconnect.

  More importantly, a PluginInstance can serve multiple sessions simultaneously: listeners are stored by sessionId, while Stop() closes the instance’s shared stdin/stdout/stderr pipes. Therefore, cancelling session A can terminate session B if both
  sessions are assigned to the same instance. IsStopped() only makes Stop() idempotent; it does not establish exclusive ownership of the instance.

  LookupSession is useful for finding the bound instance, but it does not make process-level stopping safe. The cancellation should be distinguished from normal stream closure and propagated at session level, allowing the plugin to cancel only the
  corresponding upstream request. Stopping the whole instance should only be used when the instance is guaranteed to be dedicated to that session, or as a controlled fallback.

